PR #26, opened 2019-11-26
Presspass dev

TylerFisher:
Requesting a PR into the API branch so we don't have crazy merge conflicts when trying to bring these changes into our work.


config/settings/local.py
```diff
@@ -50,3 +50,28 @@
 CELERY_ALWAYS_EAGER = False
 # Your stuff...
 # ------------------------------------------------------------------------------
+
+CORS_ORIGIN_WHITELIST = [
```
mitchelljkotler: These should be based on environment variables, not hardcoded - shouls probably also be in base instead of local

config/urls.py
```diff
@@ -27,11 +28,7 @@
     PressPassPlanViewSet,
 )
 from squarelet.users.views import LoginView
-from squarelet.users.viewsets import (
-    PressPassUserViewSet,
-    UrlAuthTokenViewSet,
-    UserViewSet,
-)
+from squarelet.users.viewsets import PressPassRegisterView, PressPassUserViewSet, UrlAuthTokenViewSet, UserViewSet
```
mitchelljkotler: Please use the supplied isort and black formatting

squarelet/auth_helpers/urls.py
```diff
@@ -0,0 +1,25 @@
...
+from django.urls import path
+
+# Third party
+from rest_auth.views import (
+    LoginView, LogoutView, PasswordChangeView,
+    PasswordResetView, PasswordResetConfirmView
+)
+
+# Local
+from . import views
```
mitchelljkotler: I've settled on preferring absolute imports to relative imports - I know the code base still has some relative imports but have been meaning to switch - am open to discussing the merits of each

squarelet/users/viewsets.py
```diff
@@ -92,6 +101,33 @@ def retrieve(self, request, pk=None):
...
+        if (
+            allauth_settings.EMAIL_VERIFICATION
+            == allauth_settings.EmailVerificationMethod.MANDATORY
+        ):
+            return {"detail": _("Verification e-mail sent.")}
+        return {}
+
+    def perform_create(self, serializer):
+        data = serializer.get_cleaned_data()
+        data["name"] = ""
```
mitchelljkotler: Why no name?

squarelet/users/viewsets.py
```diff
@@ -92,6 +101,33 @@ def retrieve(self, request, pk=None):
...
+        data["source"] = "PressPass" # TODO: Is this the proper 'source' for MixPanel?
+        data["plan"] = Plan.objects.get(slug="free")
+        user, group_organization, error = get_user_model().objects.register_user(data)
+        user.save()
+
+        print(self.request)
+        print(self.request._request)
```
mitchelljkotler: Prefer debug prints to not be checked in

squarelet/users/viewsets.py
TylerFisher: It's not included in `django-rest-auth`'s serializer, so we need to explicitly set it on the dict;

milesmcc:
The most recent six commits on `presspass-dev` hopefully solve most of the concerns left in the code review, though there are some areas (e.g. with the absolute/relative import, and how I incorporated the `CSRF_TRUSTED_ORIGINS` and `CORS_ORIGIN_WHITELIST` into `settings/base.py` that I'm not 100% sure about.

PR #27, opened 2020-01-07
Change nginx to enable dev-mode websockets to work

freedmand:
Add minimal config to pass along websocket requests


jacqui:
+1 I had to make this change too.

PR #32, opened 2020-03-02
New PressPass endpoints for memberships by user 

jacqui:
This PR adds one endpoint to the `/pp-api` that will allow us to request the memberships a user has in organizations. I added this separate endpoint as we wanted to include info on whether or not the user is an admin, which isn't included in the `/organizations?user={uuid}` response.

* `/pp-api/users/{user_uuid}/memberships/` returns a list of memberships

### DIFF
--- a/squarelet/organizations/querysets.py
+++ b/squarelet/organizations/querysets.py
@@ -48,6 +48,9 @@ class OrganizationQuerySet(models.QuerySet):
 class MembershipQuerySet(models.QuerySet):
     def get_viewable(self, user):
         """Returns memberships in public orgs or any org the user is a member of"""
+        # you can view membership info if:
+        #  * this organization is public, regardless of your membership
+        #  * or, you are a member of the org
         return self.filter(Q(organization__private=False) | Q(organization__users=user))
 
 
--- a/squarelet/organizations/querysets.py
+++ b/squarelet/organizations/querysets.py
@@ -47,7 +47,8 @@ class OrganizationQuerySet(models.QuerySet):
 
 class MembershipQuerySet(models.QuerySet):
     def get_viewable(self, user):
-        return user.memberships.filter(Q(private=False) | Q(organization__users=self.request.user))
+        """Returns memberships in either public organizations or any org the user is a member of"""
+        return self.filter(Q(organization__private=False) | Q(organization__users=user))
 
 
 class PlanQuerySet(models.QuerySet):
--- a/squarelet/organizations/querysets.py
+++ b/squarelet/organizations/querysets.py
@@ -47,7 +47,7 @@ class OrganizationQuerySet(models.QuerySet):
 
 class MembershipQuerySet(models.QuerySet):
     def get_viewable(self, user):
-        """Returns memberships in either public organizations or any org the user is a member of"""
+        """Returns memberships in public orgs or any org the user is a member of"""
         return self.filter(Q(organization__private=False) | Q(organization__users=user))
 
 
--- a/squarelet/organizations/querysets.py
+++ b/squarelet/organizations/querysets.py
@@ -45,6 +45,11 @@ class OrganizationQuerySet(models.QuerySet):
         return user.individual_organization
 
 
+class MembershipQuerySet(models.QuerySet):
+    def get_viewable(self, user):
+        return user.memberships.filter(Q(private=False) | Q(organization__users=self.request.user))
+
+
 class PlanQuerySet(models.QuerySet):
     def get_viewable(self, user):
         if user.is_staff:
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -199,7 +199,6 @@ class PressPassNestedInvitationSerializer(serializers.ModelSerializer):
             "created_at": {"read_only": True},
             "accepted_at": {"read_only": True},
             "rejected_at": {"read_only": True},
-            "uuid": {"read_only": True},
         }
 
     def validate_email(self, value):
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -273,7 +273,6 @@ class PressPassEntitlmentSerializer(serializers.ModelSerializer):
         super().__init__(*args, **kwargs)
         context = kwargs.get("context", {})
         request = context.get("request")
-        view = context.get("view")
 
         # may only create entitlements for your own clients
         if request and request.user and request.user.is_authenticated:
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -177,7 +177,6 @@ class PressPassNestedInvitationSerializer(serializers.ModelSerializer):
     class Meta:
         model = Invitation
         fields = (
-            "uuid",
             "email",
             "user",
             "request",
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -264,20 +264,22 @@ class PressPassClientSerializer(serializers.ModelSerializer):
 
 
 class PressPassEntitlmentSerializer(serializers.ModelSerializer):
-    client = serializers.SerializerMethodField()
+    client_data = serializers.SerializerMethodField()
 
-    def get_client(self, obj):
+    def get_client_data(self, obj):
         return PressPassClientSerializer(obj.client).data
 
     class Meta:
         model = Entitlement
-        fields = ("name", "slug", "client", "description")
+        fields = ("name", "slug", "client", "description", "client_data")
         extra_kwargs = {"slug": {"read_only": True}}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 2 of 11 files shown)


squarelet/organizations/serializers.py
```diff
@@ -177,6 +177,7 @@ class PressPassNestedInvitationSerializer(serializers.ModelSerializer):
     class Meta:
         model = Invitation
         fields = (
+            "uuid",
```
mitchelljkotler: Why was UUID added here?

squarelet/organizations/serializers.py
```diff
@@ -262,20 +264,22 @@ class Meta:
 
 
 class PressPassEntitlmentSerializer(serializers.ModelSerializer):
-    client = serializers.SerializerMethodField()
+    client_data = serializers.SerializerMethodField()
```
mitchelljkotler: https://www.django-rest-framework.org/api-guide/relations/#nested-relationships

This is better represented as:
```
client = PressPassClientSerializer(read_only=True)
```
I believe.  Also, why did you separate `client` and `client_data`?

squarelet/organizations/serializers.py
```diff
@@ -262,20 +264,22 @@ class Meta:
...
         model = Entitlement
-        fields = ("name", "slug", "client", "description")
+        fields = ("name", "slug", "client", "description", "client_data")
         extra_kwargs = {"slug": {"read_only": True}}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         context = kwargs.get("context", {})
         request = context.get("request")
+        view = context.get("view")
```
mitchelljkotler: It looks like you added view but did not use it.  Running pylint should catch this.

squarelet/users/serializers.py
```diff
@@ -7,8 +7,15 @@
...
 # Squarelet
-from squarelet.organizations.serializers import MembershipSerializer
+from squarelet.organizations.serializers import (
+    MembershipSerializer,
+    PressPassOrganizationSerializer
+)
 from squarelet.users.models import User
+from squarelet.organizations.models import (
+    Membership,
+    Invitation
```
mitchelljkotler: Clean up unused imports - pylint should catch these also.

squarelet/users/serializers.py
```diff
@@ -147,3 +154,15 @@ def __init__(self, *args, **kwargs):
             if self.instance and not self.instance.can_change_username:
                 # pylint: disable=invalid-sequence-index
                 self.fields["username"].read_only = True
+
+
+class PressPassUserMembershipsSerializer(serializers.ModelSerializer):
+    organization = serializers.SerializerMethodField()
+
+    def get_organization(self, obj):
+        return PressPassOrganizationSerializer(obj.organization).data
```
mitchelljkotler: Same as comments above about how to do nested relationships

squarelet/users/viewsets.py
```diff
@@ -110,6 +112,23 @@ def get_object(self):
             return super().get_object()
 
 
+class PressPassUserMembershipViewSet(
+    mixins.ListModelMixin,
+    viewsets.GenericViewSet,
+):
+    queryset = Membership.objects.none()
+    serializer_class = PressPassUserMembershipsSerializer
+    permission_classes = (DjangoObjectPermissionsOrAnonReadOnly,)
```
mitchelljkotler: Why `DjangoObjectPermissionsOrAnonReadOnly` instead of `DjangoObjectPermissions` - we generally require logging in to view user details

squarelet/users/viewsets.py
```diff
@@ -110,6 +112,23 @@ def get_object(self):
...
+    lookup_field = "user__uuid"
+
+    def get_queryset(self):
+        user = get_object_or_404(
+            User,
+            uuid=self.kwargs["user_uuid"],
+        )
+        return user.memberships.all()
```
mitchelljkotler: The membership rules for viewing (see `squarelet/organizations/rules/memberships.py` - I can explain further if its unclear how this system works) - only allow viewing a membership if either the organization is public or if you are also a member of the organization - allowing private organizations to keep their memberships private.  This would require something like:
```
user.memberships.filter(Q(private=False) | Q(organization__users=self.request.user))
```
Would want to put this into a `get_viewable` method on the queryset.  Other models follow this pattern.

squarelet/organizations/serializers.py
jacqui: Ah thanks for catching this change - I just removed it. I was trying to understand how the invitations worked for building the PP front-end and thought I needed the unique ID... but I do not.

squarelet/organizations/querysets.py
```diff
@@ -45,6 +45,11 @@ def create_individual(self, user):
         return user.individual_organization
 
 
+class MembershipQuerySet(models.QuerySet):
+    def get_viewable(self, user):
+        return user.memberships.filter(Q(private=False) | Q(organization__users=self.request.user))
```
mitchelljkotler: We should add a test to this - I see I have not done a good job at testing querysets explicitly in the past, but might be a good time to add them.

I know you took this code from my comment, but that was written in the context of a view and not a queryset, and needs to be adapted to work like this.  You may want to read up on QuerySets: https://docs.djangoproject.com/en/3.0/ref/models/querysets/

1) The `get_viewable` method should return all viewable memberships for the user, not just memberships for that user  - so it should be `return self.filter(...` not `return user.memberships.filter(...` as in the view, where we only want memberships for the user we have specified in the URL.  So in the view, with this method, it would be something along the lines of `user.memberships.get_viewable(self.request.user)`

2) `private=False` is on the Organization, not the membership, so it should be `organization__private=False`

3) `Q(organization__users=self.request.user)` should be `Q(organization__users=user)` - we want memberships that the passed in user are a part of.  `self` here is the queryset, which does not have a request attribute - that only exists at the view level.

squarelet/organizations/serializers.py
```diff
@@ -190,6 +190,7 @@ class Meta:
             "created_at": {"read_only": True},
             "accepted_at": {"read_only": True},
             "rejected_at": {"read_only": True},
+            "uuid": {"read_only": True},
```
mitchelljkotler: This should not be added

squarelet/users/serializers.py
```diff
@@ -147,3 +153,12 @@ def __init__(self, *args, **kwargs):
             if self.instance and not self.instance.can_change_username:
                 # pylint: disable=invalid-sequence-index
                 self.fields["username"].read_only = True
+
+
+class PressPassUserMembershipsSerializer(serializers.ModelSerializer):
```
mitchelljkotler: It is an arbitrary decision, but I chose `Membership` to live in the `Organization` app, so it should be in `squarelet/organizations/serializers.py`

squarelet/users/viewsets.py
```diff
@@ -110,6 +112,23 @@ def get_object(self):
             return super().get_object()
 
 
+class PressPassUserMembershipViewSet(
```
mitchelljkotler: This should also be in the organization app

squarelet/users/viewsets.py
mitchelljkotler: This should be changed to `return user.memberships.get_viewable(self.request.user)`

squarelet/organizations/querysets.py
jacqui: Got it - and I read through most of the QuerySet documentation. I also reviewed the viewset docs again so I'm clear. I made these changes, and I'll now take a stab at writing tests for it.

squarelet/organizations/serializers.py
jacqui: I added this id for the invitation to the data returned here because I didn't see another way of getting it. Our frontend needs the invitation ID in order to accept or reject it - it calls the API's `/pp-api/invitations/{uuid}` endpoint to do so. I've removed this for now, but I'm hoping we can chat about how best to accomplish this @TylerFisher, thanks!

squarelet/users/viewsets.py
jacqui: I updated this call in the viewset to use the get_viewable queryset method, thanks.

squarelet/users/serializers.py
jacqui: Okay, no problem. I've moved it to the Organization app.

squarelet/users/viewsets.py
jacqui: This has been moved as well.

squarelet/organizations/querysets.py
jacqui: I've added tests for the queryset directly (new file: test_querysets.py) and the expected api response (test_api.py)

squarelet/organizations/serializers.py
jacqui: I removed this line.

jacqui: Note that after our discussion about IDs earlier today that I've added this back :)

PR #36, opened 2020-03-09
Flex fields allows API to expand related data on demand

jacqui:
Hi all, this PR builds on my previous (still open) PR #32. I needed the changes made there, and I wasn't sure how you usually dealt with still in-progress work. 

So this is really 2 commits, one that adds [drf-flex-fields](https://github.com/rsinger86/drf-flex-fields) as a local dependency, the other that implements it for the organizations serializers. I figured it made sense to only add the `expandable_fields` declarations where I knew for sure that I needed them now instead of guessing across the board. 

I also wanted to get feedback sooner than later on how I've done this:

* does a new package go into `local.in` and `local.txt` first? 
* should I have added it somewhere else as well or instead of `local.in`?
* one final note: I couldn't get `inv pip-compile` to work so I had to run this locally `python manage.py runserver`; I'd love to know if this all works for others in docker.

### DIFF
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -19,7 +19,7 @@ from squarelet.organizations.models import (
 )
 
 
-class OrganizationSerializer(FlexFieldsModelSerializer):
+class OrganizationSerializer(serializers.ModelSerializer):
     uuid = serializers.UUIDField(required=False)
     # XXX remove plan once muckrock is updated to handle entitlements
     plan = serializers.SerializerMethodField()
@@ -60,7 +60,7 @@ class OrganizationSerializer(FlexFieldsModelSerializer):
         return obj.customer(StripeAccounts.muckrock).card_display
 
 
-class MembershipSerializer(FlexFieldsModelSerializer):
+class MembershipSerializer(serializers.ModelSerializer):
     organization = OrganizationSerializer()
 
     class Meta:
@@ -83,7 +83,7 @@ class StripeError(APIException):
     default_detail = "Stripe error"
 
 
-class ChargeSerializer(FlexFieldsModelSerializer):
+class ChargeSerializer(serializers.ModelSerializer):
     token = serializers.CharField(write_only=True, required=False)
     save_card = serializers.BooleanField(write_only=True, required=False)
     organization = serializers.SlugRelatedField(
@@ -136,7 +136,7 @@ class ChargeSerializer(FlexFieldsModelSerializer):
 # PressPass
 
 
-class PressPassOrganizationSerializer(FlexFieldsModelSerializer):
+class PressPassOrganizationSerializer(serializers.ModelSerializer):
     class Meta:
         model = Organization
         fields = (
@@ -164,7 +164,7 @@ class PressPassOrganizationSerializer(FlexFieldsModelSerializer):
         }
 
 
-class PressPassMembershipSerializer(FlexFieldsModelSerializer):
+class PressPassMembershipSerializer(serializers.ModelSerializer):
     user = serializers.SlugRelatedField(slug_field="uuid", read_only=True)
 
     class Meta:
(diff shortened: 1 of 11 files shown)


mitchelljkotler:
> does a new package go into local.in and local.txt first?
> should I have added it somewhere else as well or instead of local.in?

You never edit the `.txt` files manually - they are generated by `pip-compile` from the `.in` files.  `local.in` is for development requirements, such as things needed just for testing.  `production.in` is for things needed only in production.  Almost all dependencies should be placed only in `base.in`, which is merged into both local and production requirements.

> one final note: I couldn't get inv pip-compile to work so I had to run this locally python manage.py runserver; I'd love to know if this all works for others in docker.

`inv pip-compile` has been problematic in the past, but that should in no way affect running Django in docker.  You run `pip-compile` locally, then still run Django within Docker.  `pip-compile` just creates the `.txt` files from the `.in` files.  You then run `inv build` to rebuild the docker images. Then you restart the images and they will have the updated dependencies.

TylerFisher:
RE: the `pip-compile` issues, it appears that the Invoke task for running it attempts to run `pip-compile` in the Docker container. Should that task be rewritten to run locally?

jacqui:
Thanks for your feedback, @mitchelljkotler. I've responded below inline. 

> You never edit the `.txt` files manually - they are generated by `pip-compile` from the `.in` files. `local.in` is for development requirements, such as things needed just for testing. `production.in` is for things needed only in production. Almost all dependencies should be placed only in `base.in`, which is merged into both local and production requirements.

Got it. I moved the `drf-flex-fields` out of `local.in` and into `base.in` then ran `pip-compile` to regenerate the `*.txt` files in `requirements/`. This is now checked into this PR branch.

> 
> > one final note: I couldn't get inv pip-compile to work so I had to run this locally python manage.py runserver; I'd love to know if this all works for others in docker.
> 
> `inv pip-compile` has been problematic in the past, but that should in no way affect running Django in docker. You run `pip-compile` locally, then still run Django within Docker. `pip-compile` just creates the `.txt` files from the `.in` files. You then run `inv build` to rebuild the docker images. Then you restart the images and they will have the updated dependencies.

I had some problems with my local env for python that I think I've now resolved with @TylerFisher's help (thanks). I'm now using pip 19.3, python 3.7.2 and pyenv. I can pip-compile now (though the `inv pip-compile` task doesn't work for me still).

mitchelljkotler:
> RE: the `pip-compile` issues, it appears that the Invoke task for running it attempts to run `pip-compile` in the Docker container. Should that task be rewritten to run locally?

Ideally it would run in the container without issues, sparing the developer from needing to install it locally - however, this often fails for reasons that allude me, and the work around is to indeed run it locally.  Maybe an option to control running it in either would be prudent.

squarelet/organizations/serializers.py
```diff
@@ -17,7 +19,7 @@
 )
 
 
-class OrganizationSerializer(serializers.ModelSerializer):
+class OrganizationSerializer(FlexFieldsModelSerializer):
```
mitchelljkotler: Every serializer does not need to inherit from `FlexFieldsModelSerializer` - only the ones where you need the functionality, which should just be `PressPassEntitlmentSerializer` for now I believe

jacqui: Got it - I switched all the serializers back to inheriting from `serializers.ModelSerializer` except the two that needed expandable fields. (PP Entitlement and PP Invitation)

jacqui:
You're right about the tests, @mitchelljkotler - I added tests for all three cases where we need `expandable_fields`, then I ran the code in this branch through format & pylint and reran the test suite. All checked in now in this branch.

jacqui:
Just an FYI that I've added back the `uuid` to the `PressPassNestedInvitationSerializer` as we discussed in our API overview chat earlier today.

Issue #37, opened 2020-03-09
Registration 500's: KeyError 'organization_name'


jacqui:
I just tried to register a new account for the first time and found that it's 500ing for me with the following:

```
Internal Server Error: /pp-api/register/
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 124, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/utils/decorators.py", line 45, in _wrapper
    return bound_method(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/debug.py", line 76, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_auth/registration/views.py", line 46, in dispatch
    return super(RegisterView, self).dispatch(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/generics.py", line 192, in post
    return self.create(request, *args, **kwargs)
[...]

PR #38, opened 2020-03-12
Include plan in PressPass organization data

jacqui:
This PR has two commits:

1. my fix for the KeyError I was encountering (see issue #37)
2. a simple addition to PressPass organization data that adds the current plan info to it

### DIFF
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -135,6 +135,8 @@ class ChargeSerializer(serializers.ModelSerializer):
 
 
 class PressPassOrganizationSerializer(serializers.ModelSerializer):
+    plan = serializers.SerializerMethodField()
+
     class Meta:
         model = Organization
         fields = (
@@ -148,6 +150,7 @@ class PressPassOrganizationSerializer(serializers.ModelSerializer):
             "updated_at",
             "payment_failed",
             "avatar",
+            "plan",
         )
         extra_kwargs = {
             "uuid": {"read_only": True},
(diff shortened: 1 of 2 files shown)


squarelet/organizations/serializers.py
```diff
@@ -161,6 +164,9 @@ class Meta:
             "avatar": {"required": False},
         }
 
+    def get_plan(self, obj):
+        return obj.plan.slug if obj.plan else "free"
```
mitchelljkotler: The plan property on `Organization` is a shim which gets the MuckRock plan for the organization, as my code used to assume an Organization can only have one plan - now PressPass model has multiple plans, so adding this like this makes no sense.  Plans are linked to organizations through subscriptions, which are viewable as a subresource - `/pp-api/organizations/{uuid}/subscriptions/`

squarelet/users/managers.py
```diff
@@ -54,10 +54,9 @@ def register_user(self, user_data):
                     )
                 )
 
-            if plan and plan.for_groups:
-                group_organization = Organization.objects.create(
-                    name=user_data["organization_name"]
-                )
+            organization_name = user_data.get("organization_name")
```
mitchelljkotler: This change is masking an error somewhere else instead of fixing it - my sign up forms validates that you must supply an organization name if you sign up for a group plan - otherwise there is no organization to apply that group plan to.

Since you are not using my form, your sign up serializer should validate that they do not select a group plan if they do not supply an organization name to create.  I am not sure why you are seeing a key error here - what plan are you trying to sign for when you do that?

TylerFisher: We've been hardcoding to the default Free plan, which is marked for organizations. My guess is we should make a PressPass free plan that is marked for individuals only.

PR #40, opened 2020-03-18
Added ID to plan and subscription API responses for PressPass

jacqui:
This PR adds `id`s to the API responses for plans and subscriptions, as we discussed in our chat yesterday (the API overview discussion). I've added tests for these 4 (list/retrieve) endpoints as well to verify they return an id, and ran `inv format` on the code. The formatting made a whitespace change to the `test_querysets.py` which is the only reason why it's included here ✌️

### DIFF
--- a/squarelet/organizations/tests/test_querysets.py
+++ b/squarelet/organizations/tests/test_querysets.py
@@ -1,5 +1,7 @@
 # Third Party
 # Django
+# Django
+# Django
 from django.test import TestCase
 
 import pytest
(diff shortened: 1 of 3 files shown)


squarelet/organizations/tests/test_querysets.py
```diff
@@ -1,5 +1,7 @@
 # Third Party
 # Django
+# Django
+# Django
```
mitchelljkotler: So the `import` formatter bugs out sometimes with the comments - usually related to removing the imports for a section but leaving the comment in (looks like there were third party imports that were removed) - the empty comment seems to mess it up - deleting the rogue section comment and re-running seems to fix it.  I will fix this one but try to check to make sure it looks reasonable.

PR #43, opened 2020-03-31
User invitations endpoint

jacqui:
This PR adds a new endpoint to the API, as discussed on slack. It returns invitations for a user, complementing the existing organization-oriented invitations endpoint.

I made this endpoint data expandable on organization. There are tests for it with and without the `expand` parameter, and I ran `format` and `pylint` before committing.

`GET /pp-api/users/{user_uuid}/invitations/`

### DIFF
--- a/squarelet/users/serializers.py
+++ b/squarelet/users/serializers.py
@@ -155,33 +155,6 @@ class PressPassUserSerializer(serializers.ModelSerializer):
                 self.fields["username"].read_only = True
 
 
-class PressPassUserInvitationsSerializer(FlexFieldsModelSerializer):
-    organization = serializers.SlugRelatedField(slug_field="uuid", read_only=True)
-
-    class Meta:
-        model = Invitation
-        fields = (
-            "uuid",
-            "user",
-            "organization",
-            "request",
-            "accepted_at",
-            "rejected_at",
-            "created_at",
-        )
-        extra_kwargs = {"request": {"default": False}}
-        expandable_fields = {"organization": PressPassOrganizationSerializer}
-
-
-class PressPassUserMembershipsSerializer(serializers.ModelSerializer):
-    organization = PressPassOrganizationSerializer(read_only=True)
-
-    class Meta:
-        model = Membership
-        fields = ("organization", "admin")
-        extra_kwargs = {"admin": {"default": False}}
-
-
 class PressPassUserWriteSerializer(RegisterSerializer):
     """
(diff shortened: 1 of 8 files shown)


squarelet/users/serializers.py
```diff
@@ -154,6 +155,24 @@ def __init__(self, *args, **kwargs):
...
 
 
+class PressPassUserInvitationsSerializer(FlexFieldsModelSerializer):
+    organization = serializers.SlugRelatedField(slug_field="uuid", read_only=True)
+
+    class Meta:
+        model = Invitation
+        fields = (
+            "uuid",
+            "user",
```
mitchelljkotler, line 165: I would remove user - since this is only viewable under a given user, it should be unnecessary (similarly, the invitations nested under organization do not include it in its fields)

squarelet/users/serializers.py
```diff
@@ -154,6 +155,24 @@ def __init__(self, *args, **kwargs):
...
+            "organization",
+            "request",
+            "accepted_at",
+            "rejected_at",
+            "created_at",
+        )
+        extra_kwargs = {"request": {"default": False}}
```
mitchelljkotler, line 172: Since this is a read only view, there is no need to set a default for request

squarelet/users/viewsets.py
```diff
@@ -112,6 +113,19 @@ def get_object(self):
...
+class PressPassUserInvitationViewSet(
+    mixins.ListModelMixin, viewsets.GenericViewSet,
+):
+    queryset = Invitation.objects.none()
+    serializer_class = PressPassUserInvitationsSerializer
+    permission_classes = (DjangoObjectPermissions,)
+    lookup_field = "user__uuid"
+
+    def get_queryset(self):
+        user = get_object_or_404(User, uuid=self.kwargs["user_uuid"],)
```
mitchelljkotler, line 125: I think we should limit this to only allowing you to view your own invites - so return an empty query set unless the passed in uuid equals your own uuid.  Would also be useful to allow `me` for the uuid like in the base user view, for both this and user memberships.  That should make it safe to show all the invitations on the next line.

Also, reviewing this got me looking at the `PressPassUserMembershipViewSet` which shows all memberships for any user - it should be restricting those to memberships viewable for the current user using `user.memberships.get_viewable(self.request.user)` - you can throw that into this PR or ask @TylerFisher to make the change ( `git blame` tells me it is his code)

PR #46, opened 2020-04-02
Limit entitlements list (DRAFT)

jacqui:
This aims to limit the entitlements returned by the API to those: 

* my (public or private) organization's plans have access to
* or, I'm the (client) owner of

@TylerFisher do you mind having a look at this before I mark this as a PR ready for review? I was reaching with both my django and squarelet-data-model knowledge here 🤞

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -21,7 +21,7 @@ from squarelet.oidc.viewsets import ClientViewSet
 from squarelet.organizations.viewsets import (
     ChargeViewSet,
     OrganizationViewSet,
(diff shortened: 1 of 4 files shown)


TylerFisher:
Made a few updates -- `plans.private_organizations` represents organizations that _can_ subscribe to a plan, but doesn't necessarily indicate that they have. `plans.organizations` is accessed through the `Subscription` model, so that should capture all public and private subscriptions.

TylerFisher:
I think Github may have eaten a commit during its downtime, trying to resolve.

PR #47, opened 2020-04-02
Add query parameter to entitlements endpoint to get list of subscribed entitlements

TylerFisher:


squarelet/organizations/querysets.py
```diff
@@ -101,6 +101,15 @@ def get_viewable(self, user):
         else:
             return self.filter(plans__public=True)
 
+    def get_subscribed(self, user):
+        if user.is_authenticated:
+            return self.filter(
+                Q(plans__organizations__in=user.organizations.all())
+                | Q(client__owner=user)
```
mitchelljkotler, line 108: Does it make sense to include owned entitlements here?  I feel like maybe that should be separate?  Like one view to see all entitlements you are subscribed to and one to see all entitlements you own?

squarelet/organizations/tests/test_api.py
```diff
@@ -314,14 +314,28 @@ def test_list(self, api_client, mocker):
...
         assert len(response_json["results"]) == size
 
+    def test_subscribed(self, api_client, user):
+        """List entitlements for a subscribed user"""
+        size = 10
+        api_client.force_authenticate(user=user)
+        org = OrganizationFactory(admins=[user])
+        plan = PlanFactory(for_individuals=False, for_groups=True, public=False)
+        plan.private_organizations.add(org)
+        entitlements = EntitlementFactory.create_batch(size)
```
mitchelljkotler, line 331: You should create some entitlements which the user is not subscribed to, so the test is testing that those do *not* show up - otherwise subscribed could return extra entitlements and the test would miss it

PR #49, opened 2020-04-06
OAuth consent page styling

jacqui:
This PR adds an override to the two templates the django-oidc-provider uses for authorizing requests:

* authorize.html (shows a message along with decline and authorize buttons)
* error.html (any errors)

I discussed how to target PressPass clients with Tyler. Since this is an open-ended list of unknown client names, we figured that scoping the template to target first the set of MuckRock clients made sense. 

So here's what I did:

1. Following along with [the provider's instructions](https://django-oidc-provider.readthedocs.io/en/latest/sections/templates.html), I created the two template files in `templates/oidc_provider`
2. I copied & pasted [the default markup from the provider's page](https://django-oidc-provider.readthedocs.io/en/latest/sections/templates.html) for each 
3. I added a conditional that matches the client name against 3 known MuckRock clients: MuckRock, FOIA Machine and DocumentCloud
4. Anything else gets a modestly styled PressPass version of each page

I'm not a fan of having logic in templates, but a simple if/else seemed okay. 

Matching on the client name, though, seems like it'll have the potential for unexpected behaviours (will the MuckRock client names or set ever change?). If there's a better way to distinguish MuckRock vs PressPass clients I'm definitely all ears.


mitchelljkotler:
My preference would be to add a field to the `ClientProfile` model (in `squarelet/oidc/models.py`), which I created to extend OIDC Client's with additional information, which marks each client as belonging to MuckRock or PressPass.

TylerFisher:
We could do that. I'll take a stab at moving to that model today.

jacqui:
Hey @TylerFisher that change worked for me. I made sure to set my client's Client Profile to PressPass, and it all rendered as expected. Thanks!

PR #51, opened 2020-04-08
Override/customize email templates and sender

jacqui:
This PR includes changes to allow for customizing various emails for PressPass. While these are intended for our own purposes on the PressPass site, I think with slight modifications this approach could be used to support other sites, if desired. I kept it limited to PressPass though as I'm not familiar with the product needs beyond it :)

Main changes:

* adding a `source` parameter to the email class
* setting this `source` to `presspass` for invitations and requests - these are triggered from a presspass-specific api call
* conditionally setting this `source` to `presspass` on user registration emails based on the request host matching the (new) `PRESSPASS_URL` setting
* presspass email templates under `templates/presspass`

Before, an invitation email triggered from PressPass looked like:

Screen Shot 2020-04-08 at 2 07 35 pm

With these changes:

Screen Shot 2020-04-08 at 2 07 55 pm

We'll have to do a proper review of the email templates' design and content in a later step.


TylerFisher:
Yeah, it looks like that happens through allauth. Our registration view calls `complete_signup` from allauth, which ends up calling [`send_confirmation_mail`](https://github.com/pennersr/django-allauth/blob/4638bbc061091b5b0f8fd8ea877550ae23fe7660/allauth/account/adapter.py#L438-L455), rendering a template at an expected location of `account/messages/email_confirmation_signup`. It's a pretty circuitous path to get there. https://github.com/MuckRock/squarelet/blob/api/squarelet/users/viewsets.py#L135

config/settings/base.py
```diff
@@ -242,6 +242,9 @@
 DEFAULT_FROM_EMAIL = env(
     "DJANGO_DEFAULT_FROM_EMAIL", default="MuckRock <info@muckrock.com>"
 )
+PRESSPASS_FROM_EMAIL = env(
+    "DJANGO_DEFAULT_FROM_EMAIL", default="MuckRock <info@muckrock.com>"
```
TylerFisher, line 246: This should be something different from the default, right? We'll need to support a default (MuckRock) and PressPass in the same environment.

TylerFisher:
We're still missing a couple places where emails are sent:

- [Failed payments](https://github.com/MuckRock/squarelet/blob/a23792c921fc9e6c8a47ce6eca3cd344c920a014/squarelet/organizations/tasks.py#L137-L143)
- [Changed email address](https://github.com/MuckRock/squarelet/blob/85c707e33fba7f44151f65e43300a3d45497904e/squarelet/users/signals.py#L31-L39)

TylerFisher:
There is also [`send_receipt`](https://github.com/MuckRock/squarelet/blob/0b99954085a280211ab77e80651e91eee0398112/squarelet/organizations/models.py#L822-L834), which will need to be updated with a source. @mitchelljkotler, it seems some of this payment-oriented code is still operating with the old `customer` property on Organization rather than the new model. Which should we use?

config/settings/base.py
jacqui: doh! copy/paste (user) error. I'll fix this.

jacqui:
> We're still missing a couple places where emails are sent:
> 
> * [Failed payments](https://github.com/MuckRock/squarelet/blob/a23792c921fc9e6c8a47ce6eca3cd344c920a014/squarelet/organizations/tasks.py#L137-L143)
> * [Changed email address](https://github.com/MuckRock/squarelet/blob/85c707e33fba7f44151f65e43300a3d45497904e/squarelet/users/signals.py#L31-L39)

I've updated the email changed code to use presspass versions of the templates by passing the `user.source` value to `send_mail`. I didn't see where this was triggered in our PressPass front-end though - do we need to add a `change email address` feature @TylerFisher?

PR #52, opened 2020-04-17
API endpoints for managing email

jacqui:
This extends the API adding endpoints to manage email addresses.


mitchelljkotler:
> So far this includes the following POST requests:
>
>    /pp-api/users/email/add add a new email address, sends confirmation email; uses allauth's AddEmailForm, validating and saving it.
>    /pp-api/users/email/remove removes an email address (if it's not the primary address)
>    /pp-api/users/email/primary makes an email address primary (if it's verified)
>    /pp-api/users/email/send (re)sends the confirmation email

I would much prefer a RESTful design, something like:

`POST  /pp-api/users/me/emails/` - Create new email
`GET  /pp-api/users/me/emails/` - Show all emails for the user
`DELETE  /pp-api/users/me/emails/mitch@example.com/` - Remove email
`PATCH /pp-api/users/me/emails/mitch@example.com/` - Update email address

Updating can let you make an email address primary, and it should take care of making the current primary no longer primary.  I think updating could also be using for resending the confirmation, but an extra `/send/` route for that might also make sense since it is a little different.  Django REST Framework makes this style of API very easy and will route this one by default - you can use the other endpoints as examples.  The tricky part is calling the correct functions from AllAuth.

mitchelljkotler:
Thought about this a little more - I don't think there is a compelling reason to access emails for user accounts besides your own, so changing the URLs to:

`POST  /pp-api/emails/` - Create new email
`GET  /pp-api/emails/` - Show all emails for the user
`DELETE  /pp-api/emails/mitch@example.com/` - Remove email
`PATCH /pp-api/emails/mitch@example.com/` - Update email address

and always filtering by the current user should make things simpler.  Also, I saw you were trying to re-purpose the AllAuth email view - I think we need A Django REST Framework serializer and Viewset, and have the viewset call the AllAuth functions.

TylerFisher:
Started to build a DRF viewset. Listing and creating works, can't quite hack together the PATCH yet because AllAuth's functions are pretty wrapped up in its own view logic. For example, it expects all updates to be POST requests, and most of the logic is trapped in private methods. Still working to figure that part out.

squarelet/email_api/serializers.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class PressPassEmailAddressSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = EmailAddress
+        fields = ("email", "verified", "primary")
+
+    def validate(self, attrs):
+        request = self.context.get("request")
+        if request.data.get("action") == "update":
```
mitchelljkotler, line 13: I know you mentioned this, but yeah - definitely do not do this.

The serialize is meant to serialize and de-serialize between the API input (JSON or form encoded) and the Python model - validate should validate that the JSON is valid for the Python model.  On deletion, there is no data, just delete this instance, so the serializer should not be invoked.  I think the validation for deletion can just happen in the View and that will be fine.  Then validate can just focus on the update and create actions.

squarelet/email_api/tests/factories.py
```diff
@@ -0,0 +1,17 @@
...
+from allauth.account.models import EmailAddress
+
+# Squarelet
+from squarelet.users.tests.factories import UserFactory
+
+
+class EmailFactory(factory.django.DjangoModelFactory):
+    email = factory.Sequence(lambda n: f"user-email-{n}@example.com")
+    verified = False
+    primary = factory.Sequence(lambda n: False if n > 0 else True)
```
mitchelljkotler, line 12: This is not tied to one user, so could make things confusing - I think it might be better to just set this to False and explicitly set to true when needed

squarelet/email_api/tests/test_api.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_list(self, api_client, user):
+        """List emails for user"""
+        emails_for_user = 10
+        emails_for_other_users = 5
+        api_client.force_authenticate(user=user)
+        EmailFactory.create_batch(emails_for_user, user=user)
+        EmailFactory.create_batch(emails_for_other_users)
+        response = api_client.get(f"/pp-api/emails/")
+        assert response.status_code == status.HTTP_200_OK
+        response_json = json.loads(response.content)
```
mitchelljkotler, line 24: `response.json()` is a short cut for this

squarelet/email_api/viewsets.py
```diff
@@ -0,0 +1,122 @@
...
+from allauth.account.forms import AddEmailForm
+from allauth.account.models import EmailAddress, EmailConfirmation, EmailConfirmationHMAC
+from rest_framework import mixins, status, viewsets
+from rest_framework.response import Response
+
+# Squarelet
+from squarelet.email_api.serializers import PressPassEmailAddressSerializer
+
+
+class PressPassEmailConfirmationViewSet(
```
mitchelljkotler, line 15: For a one off endpoint like this, it makes more sense to use an `APIView` (https://www.django-rest-framework.org/api-guide/views/) then a `ViewSet` which is designed for handling multiple CRUD endpoints at once

squarelet/email_api/viewsets.py
```diff
@@ -0,0 +1,122 @@
...
+    mixins.ListModelMixin,
+    mixins.DestroyModelMixin,
+    viewsets.GenericViewSet,
+):
+    queryset = EmailAddress.objects.none()
+    lookup_field = "email"
+    lookup_value_regex = "[^/]+"
+    serializer_class = PressPassEmailAddressSerializer
+
+    def get_object(self, email):
```
mitchelljkotler, line 47: Why are you overriding `get_object`?  It should return the correct email instance based on the URL instead of passing the email address in as a parameter.  I think this depends on overriding `get_queryset` to select all email addresses for the current user, so we do not allow users to interact with other user's email addresses

squarelet/email_api/viewsets.py
```diff
@@ -0,0 +1,122 @@
...
+        return get_object_or_404(EmailAddress, email=email)
+
+    def list(self, request, user_uuid=None):
```
mitchelljkotler, line 50: I believe this would just work with a proper `get_queryset`

squarelet/email_api/viewsets.py
```diff
@@ -0,0 +1,122 @@
...
+    def update(self, request, user_uuid=None, email=None, partial=False):
+        email_address = self.get_object(email)
+        request.data["action"] = "update"
+        serializer = self.get_serializer(instance=email_address, data=request.data)
+
+        if serializer.is_valid():
+            email_address.set_as_primary()
+
+            try:
+                from_email_address = EmailAddress.objects.get(
```
mitchelljkotler, line 79: is this always going to be `email_address` since it was just set as primary above?  Does `set_as_primary` automatically un-set other primary emails?

squarelet/email_api/viewsets.py
```diff
@@ -0,0 +1,122 @@
...
+
+            # just so we can return this to the client
+            serializer = PressPassEmailAddressSerializer(email_address)
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        else:
+            return Response(
+                "Please specify a valid email confirmation key", status=status.HTTP_400_BAD_REQUEST
+            )
+
+    def update(self, request, user_uuid=None, email=None, partial=False):
```
mitchelljkotler, line 70: I think overriding `perform_update` instead of update might be cleaner here - can discuss this more if you want

squarelet/email_api/viewsets.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+
+    def destroy(self, request, user_uuid=None, email=None):
+        email_address = self.get_object(email)
+        data = {
+            "action": "delete",
+            "primary": email_address.primary,
+            "email": email_address.email
+        }
+        serializer = self.get_serializer(instance=email_address, data=data)
```
mitchelljkotler, line 107: Ditch the serializer and just validate here - i think that will just simplify this a lot

squarelet/users/signals.py
```diff
@@ -17,8 +17,13 @@ def email_confirmed(request, email_address, **kwargs):
 
 def email_changed(request, user, from_email_address, to_email_address, **kwargs):
     """The user has changed their primary email"""
-    # update their stripe customer
-    customer = user.individual_organization.customer
+    # update their stripe customer for muckrock
+    customer = user.individual_organization.customer(0).stripe_customer
```
mitchelljkotler, line 21: You should import and use the `StripeAccounts` choices here instead of hard coding 0 and 1 for muckrock and presspass

squarelet/email_api/viewsets.py
TylerFisher: Good catch, needed to do this first. Investigating this brought up that there's a method `get_primary` already on the model, so now I'm using that.

PR #55, opened 2020-04-30
Limit clients api to admin-only users

jacqui:
This PR changes who can access the `/pp-api/clients` endpoints in the API, restricting access to admins only. I've updated the tests to reflect this in general, and I've also added an additional test for each endpoint.

### DIFF
--- a/squarelet/oidc/rules.py
+++ b/squarelet/oidc/rules.py
@@ -1,7 +1,7 @@
 # pylint: disable=invalid-unary-operand-type
 
 # Third Party
-from rules import add_perm, is_staff, predicate
+from rules import add_perm, is_authenticated, is_staff, predicate
 
 # Squarelet
 from squarelet.core.rules import skip_if_not_obj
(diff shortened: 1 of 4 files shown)


squarelet/oidc/viewsets.py
```diff
@@ -19,7 +19,7 @@ class ClientViewSet(viewsets.ModelViewSet):
 
     queryset = Client.objects.none()
     serializer_class = ClientSerializer
-    permission_classes = (DjangoObjectPermissions,)
+    permission_classes = (IsAdminUser,)
```
mitchelljkotler, line 22: This should be left as `DjangoObjectPermissions` and the permissions should be controlled through the `squarelet/oidc/rules.py` file

TylerFisher:
My bad, didn't catch that.

squarelet/oidc/rules.py
```diff
@@ -13,7 +13,7 @@ def is_owner(user, client):
     return user == client.owner
 
 
-add_perm("oidc_provider.view_client", is_authenticated & is_owner)
-add_perm("oidc_provider.add_client", is_authenticated)
-add_perm("oidc_provider.change_client", is_authenticated & is_owner)
-add_perm("oidc_provider.delete_client", is_authenticated & is_owner)
+add_perm("oidc_provider.view_client", is_staff)
```
mitchelljkotler, line 16: So only staff will create clients, but do you still want users to be able to view a list of their own clients?

TylerFisher: Not at the moment, though I could imagine exposing that at some point.

PR #71, opened 2022-06-01
Add Big Local News to the site

palewire:
This draft pull request is the first step towards an eventual proposal to integrate Big Local News into the site. To start, it only includes the addition of the service to the header and footer. I suspect much more will be necessary, but I don't know exactly what. So I figured I'd start the conversation.


palewire:
I believe this has been superseded by #73

PR #207, opened 2024-10-09
Log whenever a user logs in to a client

mitchelljkotler:
Log every time someone uses Squarelet to log in to a client site (MuckRock, DocumentCloud, Big Local, etc)


squarelet/oidc/utils.py
```diff
@@ -19,3 +19,15 @@ def send_cache_invalidations(model, uuids):
         logger.info("Sending cache invalidations for: %s %s", model, uuids)
         for client_profile in ClientProfile.objects.exclude(webhook_url=""):
             tasks.send_cache_invalidation.delay(client_profile.pk, model, uuids)
+
+
+def oidc_login_hook(request, user, client):
+    """Log which client users login to"""
+    # take an arbitrary non-individual organization, since most users will have one org
+    organization = user.organizations.filter(individual=False).first()
+    plan = None if organization is None else organization.plan
```
eyeseast, line 28: Is it worth trying to figure out if a user belongs to _any_ org with a premium plan here?

eyeseast (approved):

squarelet/users/models.py
```diff
@@ -217,3 +217,38 @@ def is_hub_eligible(self):
             | Q(groups__hub_eligible=True)
             | Q(parent__hub_eligible=True)
         ).exists()
+
+
+class LoginLog(models.Model):
```
eyeseast, line 222: Do we need a plan to truncate this table at some point when we have two billion logins?

squarelet/oidc/utils.py
mitchelljkotler: Possibly?  I didn't devote too much time thinking about it as I assume the vast majority of users are only in one org.  Depending on how accurate we need to be, it might be worth doing something smarter

squarelet/users/models.py
mitchelljkotler: I think ideally we would have somewhere we can put this data where we can easily analyze it.  Is that just here and we build some visualizations on top?  Possibly try putting it in datasette?  Or use a third party tool that would let us analyze it?  I am also not sure if we want to just truncate after a certain amount of time or if we want to archive old data for historical purposes

allanlasser:
How can team members get data out for analysis?

PR #227, opened 2025-02-03
Updates Sign-In Page

### LINKED ISSUE #215: Update Sign In page
image

# Spec

The sign in page is composed from two components that share common components between them:

## 1. Overview and service list

image

This provides users with a quick summary of the value of their MuckRock account and lets them know about all the services they have access to.

Service items have four fields (_may require new model and database  migration_):

1. the service's `slug`, a machine-readable string referenced by code
1. the service's `name`, which should be a human-readable string with proper formatting
1. the service's `icon`, which should be a square PNG with a minimum resolution of 512x512
1. the service's `provider`, which is a text field
1. a short `description` of the service that enforces a 70 character maximum

The service list should be styled with `overflow-y: auto`, so that an overflowing service list scrolls independently of the form.

## 2. Sign In Form

The sign in form has two variants:

### Service-specific

image

When the `intent` field has a value that isn't `squarelet`, we should show details about the service with a corresponding slug in the sign in form's header: the icon, the name, and the provider.

### Generic

image

When the `intent` field is empty, is `squarelet`, or doesn't match an existing service, we should show a more generic sign-in form with a simpler header.

## Common Components

Common components can be copied from [`muckrock/documentcloud-frontend/src/lib/components/common`](https://github.com/MuckRock/documentcloud-frontend/tree/main/src/lib/components/common):

- `Button`, in ghost and primary modes
- `Field`
- `FieldLabel`
- `Input`

allanlasser:
Resolves #215

[Preview Deployment](https://squarelet-pi-accounts-s-mistnv.herokuapp.com/) (you'll need to be signed out)

Screenshot 2025-02-11 at 06 54 43

## Review Checklist

- [x] Code review & approval
- [x] Page copy review & approval
- [ ] Review & approval of BLN service descriptions


allanlasser:
I made the decision to merge this into our `redesign` working branch to unblock a few other pieces of development. If we get word back from partners that BLN apps need changes, we can address those before releasing the redesign features.

PR #231, opened 2025-02-07
Clone staging database into review apps

### LINKED ISSUE #225: Create staging database for each preview deployment
Goes with #224

allanlasser:
Closes #225 

This updates our Heroku Pipeline, running a `postdeploy` script that copies the staging database into the review app's database. This should only run when `HEROKU_APP_NAME` env var is present, which should only be in the context of a review app.

This allows us to deploy review apps that contain migrations, without messing up the schema of the staging database.

### DIFF
--- a/postdeploy.sh
+++ b/postdeploy.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 if [ -n "$HEROKU_APP_NAME" ]; then
-  heroku pg:copy squarelet-staging::DATABASE_URL DATABASE_URL --app $HEROKU_APP_NAME
+  heroku run "heroku pg:copy squarelet-staging::$DATABASE_URL $DATABASE_URL --app $HEROKU_APP_NAME --confirm $HEROKU_APP_NAME" -a $HEROKU_APP_NAME
 fi
--- a/postdeploy.sh
+++ b/postdeploy.sh
@@ -1,10 +1,5 @@
 #!/bin/bash
 
-echo "Build static files to be collected"
-npm ci --include dev
-npm run build
-ls -la frontend/dist # debug
-
-echo "Setup Django"
-python manage.py migrate --noinput
-python manage.py collectstatic --noinput -v 3 # verbose for debugging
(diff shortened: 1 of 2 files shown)


eyeseast:
I can't tell when migrations run in this version. I see `collectstatic` in the logs but that's it.

postdeploy.sh
```diff
@@ -1,10 +1,5 @@
 #!/bin/bash
```
mitchelljkotler, line 2: Can we get some comments on how this works.  Why would $HEROKU_APP_NAME not be defined?
Why is the heroku command being run in a nested heroku command?

mitchelljkotler (changes requested):

allanlasser:
> I can't tell when migrations run in this version. I see `collectstatic` in the logs but that's it.

You're absolutely right, I don't see migrations running either. I can add that command to the postdeploy after copying the database contents.

postdeploy.sh
allanlasser: Updated! Turned out that the nested command was unnecessary.

postdeploy.sh
```diff
@@ -1,5 +1,15 @@
...
+# This is inferred from the presence of a HEROKU_APP_NAME environment variable,
+# which is automatically injected by Heroku into review apps.
+# 
+# By checking that we're in a "staging" environment, we have confidence that we're
+# only copying staging data into an environment that expects to receive it.
+if [ -n "$HEROKU_APP_NAME" ] && [ "$DJANGO_ENV" = "staging" ]; then
+  # Copy the data from the staging app database to the review app database.
+  heroku pg:copy squarelet-staging::DATABASE_URL DATABASE_URL --app $HEROKU_APP_NAME --confirm $HEROKU_APP_NAME
+  # Migrate the review app database.
+  heroku run "python manage.py migrate" -a $HEROKU_APP_NAME
```
allanlasser, line 14: It might be unnecessary to wrap the `python manage.py migrate` command in a `heroku run` command.

eyeseast (approved):

PR #236, opened 2025-02-11
Fix Vite issues with Docker, update README

### LINKED ISSUE #184: Update README
I had to run **npm install** vite and then inv build to get squarelet working locally after a git pull. Should probably be added to documentation. Chris ran into a similar issue with having to run **inv build** because of a ModuleNotFoundError: No module named 'robots' error. I should go through with a fresh install to give our documentation a fresh look and see if I can catch any other issues that are not documented

allanlasser:
Closes #184

I ran through a fresh start, fixing errors where they popped up and adding additional context to the README. 

There were a few configuration issues with our Vite container, so I cleaned those up. It should install correctly, without requiring a local Node installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,13 @@ Set an environment variable that directs `docker compose` to use the `local.yml`
 export COMPOSE_FILE=local.yml
 ```
 
+> A command-line tool like [`direnv`](https://direnv.net/) can load this setting when you enter the project directory. With `direnv` installed, run:
+> ```bash
+> direnv allow .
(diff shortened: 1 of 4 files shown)


duckduckgrayduck (commented):
In the readme section where we list the packages that run, the names of packages are duplicated. Not sure if intentional or not
Duplicated

allanlasser:
@duckduckgrayduck Just a markdown formatting issue :) Should look better now!

duckduckgrayduck (approved):

PR #239, opened 2025-02-18
Multi-factor authentication

### LINKED ISSUE #90: 2FA

### LINKED ISSUE #238: Passkeys support
In #217, @morisy suggested we consider implementing passkeys (passwordless login using the WebAuthn API) in addition to, or instead of, two-factor authentication using time-based one-time passwords (TOTP). The good news is that [`allauth` supports WebAuthn](https://docs.allauth.org/en/dev/mfa/webauthn.html), so adopting it should be straightforward.

While it's a more secure than providing a password, [it provides a challenging user experience](https://arstechnica.com/security/2024/12/passkey-technology-is-elegant-but-its-most-definitely-not-usable-security/) and the [successful adoption of passkeys is uncertain](https://fy.blackhats.net.au/blog/2024-04-26-passkeys-a-shattered-dream/). The target audience for this feature is early adopters and the security conscious. This makes me think we can provide passkeys as an opt-in enhancement for users who are confident in setting it up, but we shouldn't rely on them (yet) as the only way users can enhance their account security.

### Prior Art

So far in my research, I've seen passkeys most often used in addition to two-factor methods.

GitHub supports both passkeys and two-factor auth:

Image

Google also provides both forms:

Image

mitchelljkotler:
Closes #90
Closes #238 

Go to `/accounts/2fa/` to activate two factor authentication for your account.  It will be enforced when you sign in.

All of the two factor pages are currently completely unstyled, but functional.


allanlasser:
@mitchelljkotler Love how simple it is to add 😄 Could you set this PR to merge into `redesign` rather than `dev`? That is the feature branch we are staging account-related work against, with the goal to release a more cohesive set of changes closer to end of February.

allanlasser:
Tested and working! I can contribute styling to this PR before we merge it into `redesign`.

duckduckgrayduck:
Also tested, and it works.

allanlasser:
Enabling passkey support was also, amazingly, trivial! This PR will now also close #238.

allanlasser:
While it was easy to enable, passkeys will require more templating and onboarding adjustments, so I've disabled the feature for now. I've got a new branch that enables them, hopefully with a very small diff to go along with it.

allanlasser:
OK, this is now fully styled and ready for another look.

duckduckgrayduck:
Everything looks good and works. Two questions:
- It currently prompts me to enable 2FA every time I login if I don't have it enabled. Do we intend on including a "Don't remind me again" button? Hitting skip only affects the current login flow, and the user is reminded again when they login. From a security perspective, we should encourage 2FA, but I just wanted to know what we think from a user experience perspective. 

-Do we plan on styling the 2FA editing from within a Squarelet profile once 2FA is already enabled? (Generate new codes, etc) or is the scope of that a different request. It looks good and is functional currently, but is different than the brand new 2FA signup template so I was curious

frontend/account/js/account.js
```diff
@@ -0,0 +1,20 @@
+(function () {
```
eyeseast, line 1: This shouldn't need to be wrapped in a function if we're using modules.

allanlasser:
Great questions!

> It currently prompts me to enable 2FA every time I login if I don't have it enabled. Do we intend on including a "Don't remind me again" button? Hitting skip only affects the current login flow, and the user is reminded again when they login. From a security perspective, we should encourage 2FA, but I just wanted to know what we think from a user experience perspective.

Nothing explicitly planned now, but definitely worth discussing the trade-offs. This pass could be iterated to only prompt again after a period of time. For example, adding a `last_2fa_prompt` field to the Users table, then check if it's been more than, say, two weeks. I'll file this as a follow-up issue.

> Do we plan on styling the 2FA editing from within a Squarelet profile once 2FA is already enabled? (Generate new codes, etc) or is the scope of that a different request. It looks good and is functional currently, but is different than the brand new 2FA signup template so I was curious.

Definitely have plans for improving how security settings are presented, but that's out of scope here. pages. This pass provides basic templating that's consistent with how we're currently presenting other account editing—it's not great, but it gets the job done. After sign-in and onboarding, we'll move onto improving user and organization. [I've got some mockups for that done, but need to flesh them out further.](https://www.figma.com/design/9KXyTVD6KfAHnXvWSbapv0/Squarelet?node-id=842-3712&t=nlfwQUQNcof0nctJ-1)

duckduckgrayduck (approved):

allanlasser:
I created #264 to continue the discussion on whether to ask to enable MFA on every sign-in.

frontend/account/js/account.js
allanlasser: This is copied wholesale from the `allauth` repo—I don't think it's getting used by any template files currently, and it's not imported into our `main.ts` entrypoint. I think it's safe to just delete it.

eyeseast (approved):

PR #252, opened 2025-03-04
Audit

mitchelljkotler:
Update packages to remove all pip-audit warnings

Tests pass locally

Should test more on staging before merging

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -107,11 +107,13 @@ AWS_DEFAULT_ACL = "public-read"
 # ------------------------
 
 AWS_S3_CUSTOM_DOMAIN = env("CLOUDFRONT_DOMAIN", default="")
+AWS_LOCATION = "static"
 
 if AWS_S3_CUSTOM_DOMAIN:
     STATIC_URL = f"https://{AWS_S3_CUSTOM_DOMAIN}/static/"
 elif CI_GIT_BRANCH:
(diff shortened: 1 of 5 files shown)


allanlasser (commented):
Code changes all look good. Probably won't have access to staging until I'm back at home.

allanlasser:
There's only one thing I'm having trouble with, which is bad URLs for staged assets in this deployed preview.

- The `STATIC_URL` is correct: `https://squarelet-staging.s3.amazonaws.com/b/audit/`
- The asset paths rendered in the template are missing the branch directory, which is the default logic for production:
```<link rel="stylesheet" href="https://squarelet-staging.s3.amazonaws.com/assets/main-BX_gBUax.css">```

This makes me think that the `django-vite` library might have gotten tripped up by something. I'm investigating further.

allanlasser:
OK, I've got a final diagnosis: something has broken with the `storages` S3 backend for staticfiles. It's not just vite-compiled assets that aren't getting the right URL, but all static files. I've rolled back all the changes I made to the `django-vite` integration.

allanlasser:
OK, I think I cracked it. `django-storages` has an env var, `AWS_LOCATION` that defaults to `""`. For some reason, the static file path generation was preferring this empty default over our `STATIC_URL` setting during path generation. Setting `AWS_LOCATION` explicitly in `settings/production.py` resolves the issue.

allanlasser (approved):
Kicked the tires and everything is working as expected.

duckduckgrayduck (approved):

PR #253, opened 2025-03-05
initial social logins

mitchelljkotler:


allanlasser:
When I go to the sign-in page in the deploy preview, it throws a 500 error.

It'd also be a good idea to base this on the `redesign` branch and target this PR against it, instead of `master`.

mitchelljkotler:
I rebased this onto the latest redesign branch.  I configured both GitHub and Google to work with the new URL.  You can login with either.  There is no styling to the new login buttons however.

duckduckgrayduck:
This works on my end. One bug: If I sign in with my google account (email sanjin@muckrock.com, which is tied to my github) first and enable 2FA, the next time I login using google it works as expected and prompts me for a 2FA code. 
If I instead go to login with my Github account(which has the same email)- it bypasses the 2FA prompt screen. It shows me the following instead, which I can click on my username to get into access. 
Bypass2FA

allanlasser (approved):

PR #255, opened 2025-03-12
Revise welcome message

duckduckgrayduck:
There are some merge conflicts from master (which revise_welcome_message is based on). This is the only file I've actually edited:
https://github.com/MuckRock/squarelet/blob/b8dca5c213f9b4a6fac52c58a12486ea7a19e651/squarelet/templates/account/email/email_confirmation_signup_message.html

### DIFF
--- a/squarelet/templates/account/email/email_confirmation_signup_message.html
+++ b/squarelet/templates/account/email/email_confirmation_signup_message.html
@@ -10,45 +10,66 @@
   {% url "account_set_password" as url_password %}
   {% autologin url_password user as url_password_ %}
   {% autologin activate_url user as activate_url_ %}
-  {% blocktrans with username=user.username %}
-    <p>
-    Thank you for creating an account on MuckRock.
-    </p>
-    <p>
-    Your username is
-    <a href="{{ url_user_ }}">
-      {{ username }}
-    </a>.
-    </p>
-    <p>
-    You will need to 
-    <a href="{{ activate_url_ }}">verify your email address</a> before you can start using our tools. 
-    You can also verify your email from your
-    <a href="{{ url_email_ }}">account email settings</a>.
-    </p>
+
+  {% blocktrans with full_name=user.full_name %}
+    <p>Welcome to your MuckRock account, {{ full_name }}!</p>
   {% endblocktrans %}
-  {% if minireg %}
+
+  {% if user.source == "squarelet" %}
     {% blocktrans %}
       <p>
-      You should also
-      <a href="{{ url_password_ }}">
-        set your password
-      </a>.
+        Your MuckRock account is your passport to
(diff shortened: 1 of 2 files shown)


allanlasser:
I merged master into `redesign` and set this PR's base branch to the `redesign` branch—now we've just got the 1 commit in the changelog!

squarelet/templates/account/email/email_confirmation_signup_message.html
```diff
@@ -10,45 +10,66 @@
...
-    <p>
-    Your username is
-    <a href="{{ url_user_ }}">
-      {{ username }}
-    </a>.
-    </p>
+
+  {% blocktrans with full_name=user.full_name %}
+    <p>Welcome to your MuckRock account, {{ full_name }}!</p>
+  {% endblocktrans %}
```
allanlasser, line 16: Should also fall back to `username` in the same way as at the bottom:
```suggestion
  {% blocktrans with full_name=user.full_name username=user.username %}
    <p>Welcome to your MuckRock account, {{ full_name|default:username }}!</p>
  {% endblocktrans %}
```

squarelet/templates/account/email/email_confirmation_signup_message.html
```diff
@@ -10,45 +10,66 @@
...
+      <p>
+        Your MuckRock account is your passport to
+        <a href="https://accounts.muckrock.com/">all the services and tools</a>
+        offered by MuckRock and our trusted partners.
+      </p>
+    {% endblocktrans %}
+  {% else %}
+    {% blocktrans with source=user.get_source_display %}
+      <p>
+        Your MuckRock account is your passport to {{ source }}
```
allanlasser, line 29: Is `source` a slug-y string, like if the user signs up for DocumentCloud will the string be `documentcloud`?

If so, we have a new app and model available on our redesign branch: `Service`. The Service model has a `slug` field which should map to the value of `source`, in addition to other display oriented fields like a proper `name`.

When creating the context for this template, we could get the Service corresponding to the user source. To access, you will likely need to rebase this branch onto the `redesign` branch (but it's up to date against `master`, so it should be clean and easy! 🤞 ).

duckduckgrayduck: It is a choice field. https://github.com/MuckRock/squarelet/blob/97484083aad1b224210ec7ecf4f64d9fa7468530/squarelet/users/models.py#L122
The line 
source=user.get_source_display
should pull the actual name and not the slug. See: https://stackoverflow.com/questions/4320679/django-display-choice-value

allanlasser (approved):

duckduckgrayduck:
Fails to pull full_name or username
Fails

PR #260, opened 2025-03-24
Squarelet invites

duckduckgrayduck:


squarelet/users/models.py
```diff
@@ -218,6 +219,32 @@ def is_hub_eligible(self):
             | Q(parent__hub_eligible=True)
         ).exists()
 
+    def get_potential_organizations(self):
+        # Get all verified email addresses for the user
+        emails = self.emailaddress_set.filter(verified=True)
+
+        # Extract the domains from the verified emails
+        domains = [email.email.split("@")[1].lower() for email in emails]
```
allanlasser, line 227: Before running this split operation, we should first filter the list for valid emails that contain a single `@` character. We can do that with a simple string test, or using Django's `django.core.validators.validate_email` function.

Since this function could let somebody access an organization's data, we should be careful making assumptions about the emails and domains we're permitting. Even though the emails are validated, it's possible for an email to contain more that one `@` character. This could be used to spoof a validated domain based on how we're matching strings.

allanlasser:
```suggestion
        emails = [email for email in emails if email.email.count('@') == 1]
        domains = [email.email.split("@")[1].lower() for email in emails]
```

squarelet/users/models.py
```diff
@@ -218,6 +219,32 @@ def is_hub_eligible(self):
...
+        """Check if the user can auto-join an organization."""
+        # Get the verified email addresses for the user
+        user_emails = self.emailaddress_set.filter(verified=True)
+        user_domains = [email.email.split("@")[1].lower() for email in user_emails]
+
+        # Get the organization's verified domains
+        org_domains = [domain.domain.lower() for domain in organization.domains.all()]
+
+        # Check if any of the user's domains match the organization's domains
+        return any(domain in org_domains for domain in user_domains)
```
allanlasser, line 246: Instead of duplicating the functionality above to get the user's verified emails and extract domains from them, we can test whether the `organization` argument is in the result from `get_potential_organizations()`. Alternatively, could we add a function like `get_user_domains` that can be tested and reused by both `get_potential_organizations` and `can_auto_join`.

allanlasser (changes requested):

Issue #267, opened 2025-03-28
Add an onboarding step for collecting payment data


allanlasser:
With our new onboarding flow, we now have a natural pathway for prompting users signing up for a professional account or organization. This could also serve as a way to collect payment info or prompt for an account upgrade when someone is signing in to access a premium service or feature, before we send them back to a service to make use of it.

Currently, we have a stripe form and plan selector in the signup form. This has the tradeoff of: on the plus side, letting someone sign up from a single webpage; on the minus side, it adds additional complexity when creating your account requires also  creating an organization/or setting up billing is a separate operation that's related to an (individual or group) organization.

The first question is: is this an idea we want to pursue, or do we want to persist plan and payment setup within our signup form?

I would advocate for moving payment and plan selection into a separate step. 

**For users,** it feels simpler to provide payment information in a focused way. If I get cold feet, don't have the information available, run into validation errors, or change my mind, I have already created an account and now have the option to figure out payment later. From a user perspective, shorter forms with more (optional) steps feels faster and less intimidating than longer forms with fewer steps.

**For us,** it provides an opportunity for deeper education on the benefits of paid accounts, provides more flexibility in when we prompt users to provide payment, gives us an opportunity to understand how users succeed or fail at the payment step, makes it easier to evolve our flow as we rethink our payments & plans, and keeps mission-critical signup logic as simple as possible.

For example, one benefit of breaking this off: when a new user wants to sign up for a professional or organization account, we can first show them the step asking them to find an organization to join, before asking for payment. They may find an existing org, and not need to (erroneously) create a new organization that will later need deduplication.

If we agree that a separate payment/plan setup step is the right move, I can provide some design and implementation specification and update our onboarding flow to incorporate it.

duckduckgrayduck:
Yea, I always found it weird that we prompt which type of account to make immediately at signup. I think separating them makes sense, and introducing easier opportunities to upgrade when it makes sense to

allanlasser:
Subscribing to a Professional plan, with reframing as an "Premium Individual" plan:

![Image](https://github.com/user-attachments/assets/271a1324-850f-4b63-a30d-257f017a02e4)

Subscribing to an Organization plan, with a reframing as an "Premium Organization" plan:

![Image](https://github.com/user-attachments/assets/e45df3f1-9ec6-4971-a987-9eeb0a56e09b)

PR #268, opened 2025-03-28
Updates Sign Up page

### LINKED ISSUE #248: Update Sign Up page
When a user doesn't have a MuckRock account, we want to present a Sign Up page that only asks them for the most necessary information for account creation.

Image

When signing up for a specific service, we highlight it:

Image

This is a sibling to #215.

allanlasser:
Closes #248 

Updates Sign Up page to bring the design inline with the new login page.

Now after signup, the user will enter the onboarding flow because the `post_login` function is called during signup. At a minimum, the user will see a nice message asking them to confirm their email address before continuing on.

This removes the Stripe fields from `SignupForm` but leaves the hidden `plan` field in place. If `plan` has a value when saving the form, the Plan slug is added to session storage for use during the onboarding flow. See #267 for more.

- Fixes a bug where the "Signup" navigation link would create incorrect URL params.

### DIFF
--- a/frontend/css/main.css
+++ b/frontend/css/main.css
@@ -1729,6 +1729,7 @@ textarea {
 }
 
 .button {
+  box-sizing: border-box;
   display: flex;
   padding: 0.375em 0.75em;
   justify-content: center;
--- a/squarelet/core/templatetags/handleintent.py
+++ b/squarelet/core/templatetags/handleintent.py
@@ -40,6 +40,7 @@ def match_service_to_intent(context):
     except Service.DoesNotExist:
         return None
 
+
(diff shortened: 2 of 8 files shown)


mitchelljkotler (commented):
This looks good.  A couple comments:
* The /selectplan/ page still exists and is the default redirect when going to the site logged out and after you log in.  Might just be out of scope for this PR, but thought it was worth noting.
* On the sign up page, should the services on the left side of the page be clickable links to their respective sites?  I feel like that would help people explore different services they are unfamiliar with.

allanlasser:
Good feedback, and worth addressing, but feels out of scope for this PR. Both can be filed as small follow-up issues!

Issue #177, opened 2024-07-24
Allow users to automatically join organizations by email domain


mitchelljkotler:
Organizations should be able to provide email domains associated with their organization, and allow users with verified email addresses with that domain to join without invitation or approval.

amandabee:
Before we can implement this we need to have a mechanism for allowing admins to explicitly say whether or not they want people to be able to join by email address.

duckduckgrayduck:
Task one is pending in https://github.com/MuckRock/squarelet/pull/269

duckduckgrayduck:
I've completed a review of verified Squarelet orgs and passed a list of domains to Mitch to add the appropriate domains to the orgs using a script/database

allanlasser:
We'll also need to display the list of verified domains for the organization within the context of the "Allow auto join" setting, so that an organization admin knows which domains they're allowing to auto-join.

amandabee:
Can we clarify the decisions that were made about which domains we did and did not set this up for? That can be described on [Internal Documentation](https://www.notion.so/muckrock/9b974f1ad5d04e7993a0e93e3d967cd4?v=7dd10139e8ba4a219c0446b3ef10705b) initially so that support is able to set and manage expectations about who will be able to use it initially and what our timeline is for making it available to more organizations.

duckduckgrayduck:
Our process is [now described in internal description](https://www.notion.so/muckrock/9b974f1ad5d04e7993a0e93e3d967cd4?v=7dd10139e8ba4a219c0446b3ef10705b&p=1d7f88926963806cbe2ae475ff82652b&pm=s).

duckduckgrayduck:
- Add the ability that verified orgs can add any email domain they want
- Create a block-list (Protonmail, Google, Yahoo, etc)
- Unverified orgs can't access this feature at all.

morisy:
Some domains for blacklist:

```
gmail.com
yahoo.com
hotmail.com
outlook.com
aol.com
icloud.com
msn.com
live.com
protonmail.com
zoho.com
gmx.com
mail.com
me.com
mac.com
comcast.net
verizon.net
att.net
sbcglobal.net
cox.net
charter.net
ymail.com
rocketmail.com
hotmail.co.uk
live.co.uk
btinternet.com
bellsouth.net
shaw.ca
earthlink.net
optonline.net
web.de
t-online.de
orange.fr
wanadoo.fr
free.fr
libero.it
virgilio.it
alice.it
mail.ru
[...]

PR #269, opened 2025-03-31
Adds allow_auto_join to automatic join flow

duckduckgrayduck:
Closes 3rd bullet point of https://github.com/MuckRock/squarelet/issues/177#issuecomment-2766970534

Updateautojoin


allanlasser:
I just saw that the failing test is due to some lousy date math. Can help fix if you need!

squarelet/organizations/tests/factories.py
```diff
@@ -15,6 +15,7 @@ class OrganizationFactory(factory.django.DjangoModelFactory):
     customer = factory.RelatedFactory(
         "squarelet.organizations.tests.factories.CustomerFactory", "organization"
     )
+    allow_auto_join = factory.LazyAttribute(lambda _: False)
```
mitchelljkotler, line 18: You don't need to use a Lazy Attribute to set it to a constant, you can just set it:
`allow_auto_join = False`

squarelet/users/tests/test_models.py
```diff
@@ -106,23 +106,26 @@ def test_can_auto_join_with_multiple_emails(user_factory):
...
+    org3 = OrganizationFactory(name="Org 3", allow_auto_join=True)
     domain3 = EmailDomainFactory.create(
         organization=org3, domain="nonmatchingdomain.com"
     )
     org3.domains.set([domain3])  # Doesn't match either email
 
     # Test that the user can auto-join organizations with matching domains
     assert user.can_auto_join(org1) is True
-    assert user.can_auto_join(org2) is True
+
```
mitchelljkotler, line 125: You don't need to assert `is True` or `is False`, you can just do:
`assert user.can_auto_join(org1)`
or
`assert not user.can_auto_join(org2)`

mitchelljkotler (changes requested):
Looks good!  A couple of small changes

PR #272, opened 2025-04-02
fix bug when modifying cancelled subscription

mitchelljkotler:
There was a bug where if you cancelled your subscription, than resubscribed before your current subscription lapsed at the end of the billing period, it would get marked as cancelled in our database and be out of sync with stripe.


duckduckgrayduck:
Is there a way to test this?

eyeseast (approved):

PR #273, opened 2025-04-02
First pass at #8 Invitation improvements

duckduckgrayduck:
1. Adds a section the a user's detail page which shows their pending invites, which they can click to reach the invite page where they can accept or reject it. 

PendingInvitations

2. Prevents an admin from sending a duplicate invite to the same email. It will show them a message to hit resend instead. It also checks if the user is already a member and prevents and invite from being sent as well. 
Tryingtosendaduplicateinvite

3. Creates a resend button under an invitation that allows an admin to resend a pending invitation. 
Resend Invitation

All tests pass locally.


squarelet/organizations/views.py
```diff
@@ -298,18 +299,47 @@ def post(self, request, *args, **kwargs):
...
+
+                existing_open_invite = Invitation.objects.filter(
+                    organization=self.organization,
+                    email__iexact=email,
+                    accepted_at__isnull=True,
+                    rejected_at__isnull=True,
+                ).exists()
+
+                if existing_open_invite:
+                    messages.warning(
```
allanlasser, line 328: Just curious about this. Why do we warn + skip if there's an existing invite, but not resend?

duckduckgrayduck: That's how I interpreted the requirements in the issue. Could change if that's the approach we want to go.

squarelet/organizations/views.py
```diff
@@ -298,18 +299,47 @@ def post(self, request, *args, **kwargs):
...
-            messages.error(request, addmember_form.errors["emails"][0])
+            messages.error(
+                request, addmember_form.errors.get("emails", ["Invalid input."])[0]
+            )
         else:
             emails = addmember_form.cleaned_data["emails"]
 
-            # Create an invitation and send it to the given email addresses
             for email in emails:
+                is_already_member = Membership.objects.filter(
```
mitchelljkotler, line 309: I feel like this new functionality would be better as a method on the Org model that can be called from the view - try to keep the views lean

squarelet/organizations/views.py
mitchelljkotler: I think resending makes sense - as long as we don't open two of the same invites

mitchelljkotler (changes requested):

squarelet/organizations/views.py
duckduckgrayduck: Done

PR #275, opened 2025-04-04
Add created_at to memberships

### LINKED ISSUE #1: Add created_at field to organization.Membership
This will let us sort users of an org by when they joined

duckduckgrayduck:
Closes #1 
If you're wondering why I chose to set created_at in save instead of using auto_add_now, it's because having it set while null=True on the migration it would still set old organizations to the current datetime. Others on Stackoverflow have noted similar behavior


mitchelljkotler:
Did you try using the `AutoCreatedField` in `squarelet.core.fields`?  That does not use `auto_add_now` as it has issues like you noted.  I think this field should allow for new fields to be Null while being consistent.

duckduckgrayduck:
> Did you try using the `AutoCreatedField` in `squarelet.core.fields`? That does not use `auto_add_now` as it has issues like you noted. I think this field should allow for new fields to be Null while being consistent.

When I use AutoCreatedField with or without the extra code to update in save, it populates the current datetime on migration. Using a datetimefield and populating on save is the only way I've gotten this to work so that the fields are Null for old values. 

Autocreatedfield

mitchelljkotler (approved):

PR #280, opened 2025-04-08
Adds list of domains under form option, let's users know how to add them. 

duckduckgrayduck:
DomainsSet
NoDomainsSet

### DIFF
--- a/squarelet/organizations/forms.py
+++ b/squarelet/organizations/forms.py
@@ -163,8 +163,9 @@ class UpdateForm(forms.ModelForm):
         label=_("Allow Auto Join"),
         required=False,
         help_text=_(
-            "Allow users to join this organization automatically"
-            " if one of their verified emails matches the organization's email domain."
+            "Allow users to join this organization without an invite"
+            "if one of their verified emails matches "
+            "one of the organization's email domains."
         ),
     )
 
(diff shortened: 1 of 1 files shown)


squarelet/organizations/forms.py
```diff
@@ -174,6 +175,26 @@ class Meta:
...
+            )
+        else:
+            self.fields["allow_auto_join"].help_text = _(
+                "<br>Allow users to join this organization without requesting "
+                "an invite if one of their verified emails matches one of the "
+                " organization's email domains. This organization does not "
+                "currently have any email domains set. "
+                "To add one, please contact our support team at "
+                "<b>info@muckrock.com</b>."
+            )
```
allanlasser, line 197: This looks like something that would be better suited to a template, where we can render the `domain_list` as an unordered list using a `{% for %}` loop.

That might require defining a new template in the `Update` view—I don't really see an "admin" or "update" template that already exists.
 https://github.com/MuckRock/squarelet/blob/9322f7162dd3a9a907853fa851f659b3e18e3aee/squarelet/organizations/views.py#L215-L216

duckduckgrayduck: I don't see one that exists. Considering that the vast majority of orgs would only have one or no verified email domains and this page likely needs more work- is this a blocker?

allanlasser: Since this page may be short lived, I don't think it's a blocker.

mitchelljkotler (approved):

PR #283, opened 2025-04-11
Merge orgs

mitchelljkotler:
https://squarelet-pi-merge-orgs-mmcozw.herokuapp.com/organizations/~merge

Merging orgs will set all FKs pointing to the org you are merging in and point them to the org you are merging in to (except for Org Logs which will point to the old org so the logs reflect which org the action happened on).  It will store the org that was merged, the timestamp and the user who merged it on the org that was merged in to.  The changes will propagate to MuckRock and DocumentCloud.

To test, create some test orgs, add some users, and merge them together.


allanlasser (commented):
Overall this looks good to me.

I will jump in and add a commit with some styling and UI changes, I think that'll be faster and easier than trying to describe and translate them.

squarelet/templates/organizations/organization_merge.html
```diff
@@ -0,0 +1,236 @@
...
+            <tr>
+              <th>Parent</th>
+              <td>{{ good.parent }}</td>
+              <td>{{ bad.parent }}</td>
+            </tr>
+            <tr>
+              <th>Charges</th>
+              <td></td>
+              <td></td>
+            </tr>
```
allanlasser, line 57: No charges provided in table—either remove row, or display charges.

Issue #264, opened 2025-03-26
"Snooze" onboarding prompts to enable MFA


allanlasser:
With #239, we ask users to enable MFA every time they sign in, while affording the option to skip it.

Right now we use a simple heuristic: do they not have MFA enabled, it not the user's very first login, and do they have a verified email address?  Both @eyeseast and @duckduckgrayduck asked whether we want to be prompting on every sign in, or hitting "snooze" for a period of time. 

The biggest question is whether it's a worthwhile tradeoff between account security and ease-of-use to not encourage users to secure their accounts. After all, the most preferable way to remove the prompt is to set up MFA!

If we decide to snooze the prompt, we'd need to persist some information into the database. I think the best approach for this would be a new column field on the User model, `last_mfa_prompt`. If a user hasn't been prompted, it would be set to the current date when entering the onboarding step for `mfa_opt_in`. The next time the user signs in, we would check that value when selecting the next onboarding step; if it's within the last `MFA_PROMPT_SNOOZE_DURATION` (let's say that is  two weeks, or the same as our cookie TTL), we would skip MFA step entirely.

mitchelljkotler:
I would lean toward not prompting every single time

eyeseast:
I think this is the right approach, and we can tweak the snooze length depending on what we see. The main situation I want to account for is the impatient journalist on deadline who just needs to post a document quickly without going through onboarding steps we've just added. We _do_ want to bug them to get MFA set up, and soon, but it can be _tomorrow_.

PR #289, opened 2025-05-06
Snooze for 2 weeks

duckduckgrayduck:
You can test by running locally and changing the snooze duration to 1 minute or something like that.


duckduckgrayduck:
Closes #264

squarelet/users/constants.py
```diff
@@ -0,0 +1,4 @@
+# Standard Library
+from datetime import timedelta
+
+MFA_PROMPT_SNOOZE_DURATION = timedelta(weeks=2)
```
mitchelljkotler, line 4: I would make this a setting in `settings.py` so it can be updated via environment variables

mitchelljkotler (changes requested):

allanlasser (approved):

Issue #241, opened 2025-02-21
Update Python runtime


allanlasser:
When deploying in Heroku, we get this warning:

```
-----> Using Python 3.10.5 specified in runtime.txt
-----> Installing Python 3.10.5
 !     Warning: A Python security update is available!
 !     
 !     Upgrade as soon as possible to: Python 3.10.16
 !     See: https://devcenter.heroku.com/articles/python-runtimes
```

duckduckgrayduck:
3.10 will be supported on Heroku until October 2026:
 https://devcenter.heroku.com/articles/python-support#supported-python-versions
When we upgrade to 3.11 we should also try upgrading to Heroku Stack 24 as we are using 22 and it won't be supported after April 2027:
https://devcenter.heroku.com/articles/heroku-22-stack#support-period

I've created a separate issue for this here:
https://github.com/MuckRock/squarelet/issues/294
and thrown it in the backlog.

PR #292, opened 2025-05-09
Don't specify minor version

duckduckgrayduck:
After discussion with Mitch, easiest approach is to just not specify minor version to get us up to the appropriate version of 3.10 since we have no other reason to update to 3.11 at this time


duckduckgrayduck:
If all works fine, as it should, would close https://github.com/MuckRock/squarelet/issues/241

duckduckgrayduck:
I'm getting  Payment error: You cannot use a Stripe token more than once: tok_0RMwswtU5ovL4Hc4Vs7HunGt. Not sure if related or not. 
Also ran into https://muckrock.sentry.io/issues/6596387462/?alert_rule_id=472594&alert_timestamp=1746820124097&alert_type=email&environment=production&notification_uuid=376db5e9-d663-43c3-aeed-b8a073934d68&project=1247105
when trying to change full name. It sent me to a 500, but then when I refreshed it seems to have updated the name successfully.

mitchelljkotler (approved):

PR #293, opened 2025-05-09
Polish services list styling

allanlasser:
- Services list now adapts to the height of the login/signup step it's presented alongside. This will enable overflow scrolling on most pages when we have a longer list of services.
- Services now link out when they have a `base_url`.
- Typography and layout refinements.

### Test & review
- **[Sign Up](https://squarelet-pi-allanlasse-qqivzh.herokuapp.com/accounts/signup/?intent=squarelet)**: note how the services list is shorter than the page's form
- **[Sign In](https://squarelet-pi-allanlasse-qqivzh.herokuapp.com/accounts/login/?intent=documentcloud)**: note how the services list is longer than the page's form

### Questions

1. Should services link to their `base_url` or their sign-in intent?

### DIFF
--- a/frontend/css/main.css
+++ b/frontend/css/main.css
@@ -1628,19 +1628,25 @@ textarea {
   line-height: normal;
 }
 
-.service-providers {
-  width: 100%;
-  margin: 3rem 0;
-  display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
-  gap: 3rem;
-}
-
 .service-provider {
   display: flex;
   flex-direction: column;
   gap: 1rem;
   align-items: flex-start;
+  padding: .75rem;
+}
+
+a.service-provider {
+  color: currentColor;
+  text-decoration: none;
+  border-radius: 0.5rem;
+  transition: background 0.2s ease;
+  background: transparent;
+}
+
+a.service-provider:hover, a.service-provider:focus {
+  background: rgba(255, 255, 255, .15);
+  text-shadow: 0px 1px 4px var(--shadow, rgba(30, 48, 56, 0.15));
 }
 
 .service-provider * {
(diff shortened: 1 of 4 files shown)


mitchelljkotler:
> The overflow-scroll is a little weird, especially on mobile when everything is stacked. Is there anything in the footer that's so important we can't push it down if the services list gets longer? (And if it is important, should we move it up?)

I agree - I think just pushing down the footer would look better

amandabee:
1. We should confirm that we are testing for each password requirement. 
     a. I was able to create an account with "Test Human" as the full name and "testhuman" as the password, so we don't really test for similarity. It seems like it does catch matches between email and password and username and password. 
2. The username error reads: "Enter a valid username. This value may contain only English letters, numbers, and ./-/_ characters." -- 
   a. I think it should say "Please enter a valid username, " and 
   b. I'm not sure about the phrase "English letters" -- I think the "Latin alphabet" is more accurate. 
   c. Does `./-/_` include `/` or not? If not, can we make that more clear? 

3. On the sign in page, I think that we should just push the footer down. 
4. The services should link to their base URL.

allanlasser:
> > The overflow-scroll is a little weird, especially on mobile when everything is stacked. Is there anything in the footer that's so important we can't push it down if the services list gets longer? (And if it is important, should we move it up?)
> 
> I agree - I think just pushing down the footer would look better

I'm fine making this change given the consensus, but two points of reponse:

1. I think it's wrong to think that there's nothing important in the footer. Most important, it contains a donate button. In addition, it contains details about what MuckRock is and how it impacts users; for example, Terms of Service and Privacy Policy.
2. I agree the overflow scrolling is a little weird when we have 5 services in the list. I'm thinking ahead to when we have 15.

allanlasser:
> * We should confirm that we are testing for each password requirement.
>   a. I was able to create an account with "Test Human" as the full name and "testhuman" as the password, so we don't really test for similarity. It seems like it does catch matches between email and password and username and password.
> * The username error reads: "Enter a valid username. This value may contain only English letters, numbers, and ./-/_ characters." --
>   a. I think it should say "Please enter a valid username, " and
>   b. I'm not sure about the phrase "English letters" -- I think the "Latin alphabet" is more accurate.
>   c. Does `./-/_` include `/` or not? If not, can we make that more clear?

This is good feedback, but out of scope for these changes. I'll open a new issue to clarify the sign up form's validation and error messages.

PR #310, opened 2025-05-15
fix stripe token double use

mitchelljkotler:
Fix a bug where the stripe token is used twice, which will throw an error


eyeseast (approved):

duckduckgrayduck:
One of the tests failed, probably needs changing

mitchelljkotler:
Fixed the failing test

Issue #295, opened 2025-05-12
API endpoints for organization management


allanlasser:

duckduckgrayduck:
Admin should view both organizations they have sent for the orgs they are admins of (request=False) and users who have requested to join their org(request=True). For invitations they have sent out- they can only revoke the invitation. 
For invitations that were sent to them (request=True), they can accept/reject. Revoke also rejects, but it is important to make this distinction. 

For the regular user- they can see both invitations they have received (by email, request=False) which they can accept/reject and requests for invites they have sent (request=True), which they can only revoke(withdraw). They cannot accept their own request to join. 

List() will automatically filter based only on the invitations that have not yet been accepted or rejected (thus they are open).

The endpoint will only be visible on /api/ for staff users

duckduckgrayduck:
Kicking this back to shaping

PR #316, opened 2025-05-20
Squarelet api changes

duckduckgrayduck:
Before I spend some time writing tests for this, I want to see if this looks good so far. Under the hood a revoke is just a reject, but I wanted to provide a different response for a revoke to assist in debugging in the future. This PR has changes that would close out everything in https://github.com/MuckRock/squarelet/issues/295
except 
- List invited members. This can be added in by finding user IDs by their associated emails from the invites, but in the case of a user that isn't yet created I wasn't sure how to handle this. 

Here is some further background for my thought process here

Admin should view both invitations they have sent for the orgs they are admins of (request=False) and users who have requested to join their org(request=True). For invitations they have sent out- they can only revoke the invitation. It doesn't make sense to me to call it a "reject" because it isn't something they are responding too, even though they are functionally the same. For invitations that were sent to the admin (request=True), they can accept/reject. As well as any requests to join  a different org they aren't an admin of. 

For the regular user- they can see both invitations they have received (by email, request=False) which they can accept/reject and requests for invites they have sent (request=True), which they can only revoke(withdraw). They cannot accept their own request to join.

List() will automatically filter based only on the invitations that have not yet been accepted or rejected (thus they are open).

The endpoint will only be visible on /api/ for staff users

### DIFF
--- a/squarelet/organizations/viewsets.py
+++ b/squarelet/organizations/viewsets.py
@@ -1,17 +1,27 @@
 # Third Party
+# Django
 from django.db.models import Q
+
 import django_filters
-from rest_framework import viewsets, filters
+from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework import filters, viewsets, status
 from rest_framework.decorators import action
 from rest_framework.permissions import IsAdminUser
 from rest_framework.response import Response
-from django_filters.rest_framework import DjangoFilterBackend
 
 # Squarelet
 from squarelet.oidc.permissions import ScopePermission
-from squarelet.organizations.models import Charge, Organization, Invitation
-from squarelet.organizations.serializers import ChargeSerializer, OrganizationSerializer, InvitationSerializer
-from squarelet.organizations.permissions import CanRevokeInvitation, CanAcceptOrRejectInvitation
+from squarelet.organizations.models import Charge, Invitation, Organization
+from squarelet.organizations.permissions import (
+    CanAcceptOrRejectInvitation,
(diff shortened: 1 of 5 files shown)


squarelet/organizations/viewsets.py
```diff
@@ -26,3 +58,66 @@ class ChargeViewSet(viewsets.ModelViewSet):
...
+    )
+    def reject(self, request):
+        invitation = self.get_object()
+        try:
+            invitation.reject()
+            return Response({"status": "invitation rejected"})
+        except ValueError as error:
+            return Response({"detail": str(error)}, status=status.HTTP_400_BAD_REQUEST)
+
+    @action(detail=True, methods=["post"], permission_classes=[CanRevokeInvitation])
```
allanlasser, line 108: Not trying to nitpick here, but I prefer "withdraw" to "revoke" as the verb we use here, for two reasons:

1. Revoke has a slightly negative tone to me, as if the recipient of the invitation or request did something wrong; "withdraw" sounds more neutral.
2. "revoke" is very similar to "reject" in sound and spelling. Having the three verbs very clearly identifiable "accept, reject, withdraw" makes it easier to recognize each at a quick glance or in conversation.

allanlasser (commented):
This looks good to me! I think this is ready for testing.

squarelet/organizations/viewsets.py
duckduckgrayduck: That sounds good to me

PR #320, opened 2025-05-21
Add TOS check, tighten similarity check

### LINKED ISSUE #309: Improve validation and feedback on Sign Up form
Feedback from @amandabee in #293:

> 1. We should confirm that we are testing for each password requirement.
>     a. I was able to create an account with "Test Human" as the full name and "testhuman" as the password, so we don't really test for similarity. It seems like it does catch matches between email and password and username and password.
> 2. The username error reads: "Enter a valid username. This value may contain only English letters, numbers, and ./-/_ characters." 
>     a. I think it should say "Please enter a valid username, " and
>     b. I'm not sure about the phrase "English letters" -- I think the "Latin alphabet" is more accurate.
>     c. Does ./-/_ include / or not? If not, can we make that more clear?

eyeseast:
Closes #309 

This is actually removing one of our password validators, so we should decide if this is what we want. Alternately, we can make it check against the full name field.

The TOS field is only on the signup form. We don't store anything, but anyone who creates an account will have to check the box that they agree to it.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -139,7 +139,7 @@ AUTH_PASSWORD_VALIDATORS = [
         "UserAttributeSimilarityValidator",
         "OPTIONS": {
             "user_attributes": ["username", "name", "email"],
-            "max_similarity": 0.5,
+            "max_similarity": 0.6,
         },
     },
     {"NAME": "django.contrib.auth.password_validation.MinimumLengthValidator"},
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -134,6 +134,14 @@ PASSWORD_HASHERS = [
 ]
 # https://docs.djangoproject.com/en/dev/ref/settings/#auth-password-validators
 AUTH_PASSWORD_VALIDATORS = [
+    {
+        "NAME": "django.contrib.auth.password_validation."
+        "UserAttributeSimilarityValidator",
+        "OPTIONS": {
+            "user_attributes": ["username", "name", "email"],
+            "max_similarity": 0.5,
+        },
+    },
     {"NAME": "django.contrib.auth.password_validation.MinimumLengthValidator"},
     {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"},
     {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator"},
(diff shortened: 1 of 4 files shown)


config/settings/base.py
```diff
@@ -134,11 +134,6 @@
 ]
 # https://docs.djangoproject.com/en/dev/ref/settings/#auth-password-validators
 AUTH_PASSWORD_VALIDATORS = [
-    {
-        "NAME": "django.contrib.auth.password_validation."
-        "UserAttributeSimilarityValidator",
```
allanlasser, line 139: Would it be possible to subclass the [`UserAttributeSimilarityValidator`](https://github.com/django/django/blob/8c279113864c31cef539c6384a683e6ed832adca/django/contrib/auth/password_validation.py#L183-L213) to ensure it checks against the user's name to our satisfaction? Based on how it's used here, it seems as though the `max_similarity` threshold could be lowered to match Amanda's expectations.

eyeseast: We were including the name, email and username fields, with `max_similarity` set to `0.7`, and we can configure those options in settings without subclassing. I guess we'd want to turn the similarity down.

eyeseast: I turned it down to `0.6`. At `0.5`, it was saying `squarelet` was too similar to `doe@example.com`.

squarelet/templates/account/signup.html
```diff
@@ -136,6 +136,17 @@
             {{ form.password1.help_text }}
           </div>
         </label>
+        <label class="field">
+          <span class="label">
+          {{ form.tos }}
+          {{ form.tos.label }}
```
allanlasser, line 142: This template already has some custom HTML in it already; I don't think we have to be too strict about using every field's programmed values:

```suggestion
          <span>I accept MuckRock&rsquo;s <a target="_blank" rel="noopener noreferer" href="https://www.muckrock.com/tos/">Terms of Service</a> and <a target="_blank" rel="noopener noreferer" href="https://www.muckrock.com/privacy-policy/">Privacy Policy</a>.</span>
```

With this, we can remove the help text and simplify the field down to one line.

allanlasser (approved):
Thanks for making those changes. Looks great 👍

PR #325, opened 2025-05-23
two factor protect password reset form

### LINKED ISSUE #306: Enforce 2FA code on password reset form
From @duckduckgrayduck:

> Something I have noticed that is possible, but is something we should reconsider, is that an attacker (who has control of an email address) can send a password reset email and even change the password, without presenting a 2FA token.There are some different thoughts about approach to this:  https://security.stackexchange.com/questions/231454/how-does-password-reset-work-if-2fa-is-enabled
but it is generally accepted best practice to allow someone to send the password reset email, but to actually change the password once they click the link they need to present a token. 
> 
> The reason why this is considered best practice is that the account can be left in a temporary indeterminate state where the attacker cannot login nor can the legitimate user. The legitimate user can’t log in because they no longer have the correct password (they would need to get a password reset request) and the attacker cannot access because it still doesn’t permit them to login with just the newly created password. This isn’t a dealbreaker for release per se since we don’t let the attacker login without a code anyway, but we should at least discuss this if we haven’t already.

mitchelljkotler:
This closes #306

### DIFF
--- a/squarelet/users/tests/test_forms.py
+++ b/squarelet/users/tests/test_forms.py
@@ -124,12 +124,17 @@ def test_save_org(rf, plan_factory, organization_plan_factory, mocker):
         (forms.ChangePasswordForm, 3),
         (forms.SetPasswordForm, 2),
         (forms.ResetPasswordForm, 1),
-        (forms.ResetPasswordKeyForm, 2),
     ],
 )
 def test_other_forms(form_class, number_fields):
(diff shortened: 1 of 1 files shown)


mitchelljkotler:
> This works, it looks like one test fails

I have fixed the failing test

duckduckgrayduck (approved):

PR #329, opened 2025-05-23
upgrade django debug toolbar

mitchelljkotler:
Upgrades django-debug-toolbar


duckduckgrayduck:
All styling seems to be dropped with this update. 
Styling

allanlasser (approved):

Issue #274, opened 2025-04-03
Users can have undeliverable email address


allanlasser:
After we've validated a user's email, it may later become undeliverable. We currently log undeliverable email addresses, so we can prompt the user to correct this:

- when signing in
- from their account page

allanlasser:
To research: how many accounts have an undeliverable email now?

If your address is undeliverable, when you log in you should get a prompt to fix it by adding and confirming a new email address.  "Undeliverable" will only happen for primary addresses, since they're the only ones we send emails to.

For secondary emails, we should send an annual email to test that the email is still deliverable. It should be a reminder of all the cool stuff you can do with your account!

duckduckgrayduck:
This already exists

PR #338, opened 2025-05-28
Prompt users to update their undeliverable primary email addresses

duckduckgrayduck:
PleaseUpdate

Part of fixing #274 

Currently we have 150 users with this issue.


mitchelljkotler:
https://github.com/MuckRock/squarelet/blob/master/squarelet/templates/base.html#L86

It should show a notification on every page when you have a bad email address already

allanlasser:
I think we can close it—I didn't realize this feature was present when filing the issue.

PR #356, opened 2025-07-02
Social login flag

mitchelljkotler:
Instead of automatically creating an account on social login, this confirms with the user if they want to log in to an existing account or create a new one.

@allanlasser - This probably needs some design tweaks

I used the copy supplied by @amandabee , but it can be further edited if needed

To test, attempt to use the social login with an email address that is not associated with an account.

https://squarelet-pi-social-log-dmdlac.herokuapp.com/accounts/login/

### DIFF
--- a/compose/local/django/start
+++ b/compose/local/django/start
@@ -6,4 +6,4 @@ set -o nounset
 
 
 python manage.py migrate
-python manage.py runserver_plus 0.0.0.0:80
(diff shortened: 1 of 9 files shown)


allanlasser:
I touched up the styling and copy on the confirmation page:

Screenshot 2025-07-02 at 13 40 34

PR #364, opened 2025-07-21
Adds filters to `/api/organization`

allanlasser:
As a client, I want to be able to filter the list of organizations to distinguish between private/public orgs, and verified/unverified orgs. This adds simple filters to the list endpoint to narrow down the returned orgs.


allanlasser:
I've found that removing entitlements from the list response is crashing MuckRock when signing in locally. I'm investigating whether I can use the `OrganizationDetailSerializer` in the list of organizations returned with a user.

PR #365, opened 2025-07-21
Added instructions if user doesn't have a code

### LINKED ISSUE #362: Include recovery code instructions on the 2FA code entry form
It's unclear to users that they can provide recovery codes as equivalent values in the authentication code input. Users benefit from some language guiding them through what to do if they can't provide a code.

Something like:

> **Don't have a code?**
> If can't provide an authentication code from your second device, you can provide a recovery code in its place. Recovery codes are one-time use and will not work on subsequent logins. [Learn more](https://help.muckrock.com/Two-Factor-Authentication-1f9f8892696380e59857c19341a324af?source=copy_link#207f889269638028a939c82281c5c6f3)

duckduckgrayduck:
Close #362 

It would be nice to have the paragraph under Don't have a code? be hidden unless a user clicked on it. If we would like that approach, it would be helpful to me to know where we would like to add Javascript

### DIFF
--- a/squarelet/templates/account/onboarding/mfa_confirm.html
+++ b/squarelet/templates/account/onboarding/mfa_confirm.html
@@ -1,5 +1,4 @@
 {% extends "account/onboarding/base.html" %}
-
 {% load i18n static %}
 {% load account socialaccount %}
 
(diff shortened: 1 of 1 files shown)


allanlasser (approved):
I applied a few changes, mostly to avoid inline styling and use consistent language.

I think it's better to display the full message instead of hiding it inside a toggle or dialog.

PR #367, opened 2025-07-21
Private frontend API

mitchelljkotler:


squarelet/organizations/fe_api/serializers.py
```diff
@@ -0,0 +1,129 @@
...
+from rest_framework import serializers, status
+from rest_framework.exceptions import APIException
+
+# Squarelet
+from squarelet.organizations.models import Invitation, Membership, Organization
+
+
+class OrganizationSerializer(serializers.ModelSerializer):
+    uuid = serializers.UUIDField(required=False)
+    # remove plan once all clients are updated to handle entitlements
```
mitchelljkotler, line 15: I think you can remove this comment - if we want the plan name to display, we can just leave it in this API

squarelet/organizations/fe_api/serializers.py
```diff
@@ -0,0 +1,129 @@
...
+            "updated_at",
+            "payment_failed",
+            "avatar_url",
+            "merged",
+        )
+
+    def get_plan(self, obj):
+        return obj.plan.slug if obj.plan else "free"
+
+    def get_entitlements(self, obj):
```
mitchelljkotler, line 48: This is going to be slow and is potentially more useful in the other API - may be a candidate for removal

squarelet/organizations/fe_api/serializers.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if client:
+            return list(
+                client.entitlements.filter(plans__organizations=obj)
+                .annotate(update_on=F("plans__subscriptions__update_on"))
+                .values("name", "slug", "description", "resources", "update_on")
+            )
+        return []
+
+    def get_member_count(self, obj):
```
mitchelljkotler, line 63: This is potentially going to be slow, we may need to find a way to optimize this if we really need it

squarelet/organizations/fe_api/serializers.py
```diff
@@ -0,0 +1,129 @@
...
+        return obj.memberships.count()
+
+    def get_admins(self, obj):
+        return list(
+            obj.memberships.filter(admin=True).values_list("user_id", flat=True)
+        )
+
+    def get_members(self, obj):
```
mitchelljkotler, line 71: DRF can handle M2M relationships natively, which should offer better performance: (for members and admins)
https://www.django-rest-framework.org/api-guide/relations/#primarykeyrelatedfield

squarelet/organizations/fe_api/serializers.py
```diff
@@ -0,0 +1,129 @@
...
+        # Remove admins and member_count if organization is individual
+        if instance.individual:
+            rep.pop("admins", None)
+            rep.pop("member_count", None)
+
+        # Only keep members if user is authenticated AND member of org
+        if not user or not user.is_authenticated:
+            rep.pop("members", None)
+        else:
+            is_member = instance.memberships.filter(user=user).exists()
```
mitchelljkotler, line 89: This is going to cause DB calls per instance and have performance issues.  If we are mainly using this for dynamic org selection, maybe we can leave members out entirely for now?

squarelet/organizations/fe_api/serializers.py
```diff
@@ -0,0 +1,129 @@
...
+            if not is_member:
+                rep.pop("members", None)
+
+        return rep
+
+
+class MembershipSerializer(serializers.ModelSerializer):
```
mitchelljkotler, line 96: I don't think this is being used anywhere

squarelet/organizations/fe_api/viewsets.py
```diff
@@ -0,0 +1,109 @@
+# Third Party
```
mitchelljkotler, line 1: remove mistaken `isort` comment

squarelet/organizations/fe_api/viewsets.py
```diff
@@ -0,0 +1,109 @@
...
+                | Q(email__in=verified_emails)
+                | Q(
+                    organization__memberships__user=user,
+                    organization__memberships__admin=True,
+                )
+            )
+            .distinct()
+        )
+
+    @action(detail=True, methods=["post"], permission_classes=[CanAcceptInvitation])
```
mitchelljkotler, line 87: Accept, reject and resend were suppose to be under a single `update` operation with an `action` field to specify what action to take

squarelet/users/fe_api/serializers.py
```diff
@@ -0,0 +1,96 @@
+# Third Party
+from allauth.socialaccount.models import SocialAccount, SocialToken
+from rest_framework import serializers
+
+# Squarelet
+from squarelet.organizations.serializers import MembershipSerializer
+from squarelet.users.models import User
+
+
+class SocialTokenSerializer(serializers.ModelSerializer):
```
mitchelljkotler, line 10: I don't think we need social tokens or accounts in this API

squarelet/users/fe_api/viewsets.py
```diff
@@ -0,0 +1,36 @@
...
+        ),
+        Prefetch(
+            "emailaddress_set",
+            queryset=EmailAddress.objects.filter(primary=True),
+            to_attr="primary_emails",
+        ),
+    ).order_by("created_at")
+    serializer_class = UserSerializer
+    permission_classes = (IsAuthenticated,)
+    lookup_field = "individual_organization_id"
```
mitchelljkotler, line 28: We can just use `id` for this API

squarelet/users/fe_api/serializers.py
```diff
@@ -0,0 +1,96 @@
...
+            "organizations",
+            "bio",
+            "picture",
+            "updated_at",
+            "use_autologin",
+            "social_accounts",
+        )
+        read_only_fields = fields
+
+    def get_primary_email_field(self, obj, field, default):
```
mitchelljkotler, line 54: A lot of this is going to be slow - I think starting simple is going to be beneficial here.  We can just use the `email` field directly for email, then we lose `email_verified` but I think that is ok.

squarelet/users/fe_api/serializers.py
```diff
@@ -0,0 +1,96 @@
...
+                return getattr(email, field, default)
+        return default
+
+    def get_email_verified(self, obj):
+        return self.get_primary_email_field(obj, "verified", False)
+
+    def get_email(self, obj):
+        return self.get_primary_email_field(obj, "email", "")
+
+    def get_organizations(self, obj):
```
mitchelljkotler, line 70: This can use the M2M support as above:
https://www.django-rest-framework.org/api-guide/relations/#primarykeyrelatedfield

squarelet/users/fe_api/viewsets.py
```diff
@@ -0,0 +1,36 @@
...
+    swagger_schema = None
+
+    def get_queryset(self):
+        # Use the get_viewable() method from MembershipQuerySet
```
mitchelljkotler, line 32: I am not sure why you are using the `Membership` `get_viewable` here.  I believe we want to show all users to everyone?  If not, we should make a `User` `get_viewable` method with the proper logic.

squarelet/users/fe_api/serializers.py
```diff
@@ -0,0 +1,96 @@
...
+        # Return list of organization IDs user is a member of
+        return [m.organization_id for m in obj.memberships.all()]
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        request = self.context.get("request")
+
+        if request and request.user.is_authenticated:
+            # Fetch the requesting user's orgs
+            requester_org_ids = set(
```
mitchelljkotler, line 80: This is going to generate DB calls per instance as well.  Maybe we omit emails entirely in the list view or do something else here.

mitchelljkotler (commented):
Solid start, but I don't think we need to stink as closely to the old API as this one does.  We can start with the minimum fields to ensure high performance, and add on as needed, with proper optimizations.  The old API did not optimize for list operations, as I only used the detail views - I will need to revisit them to optimize the list operations.

For these, we want to avoid making DB calls per object shown in the API.  The easiest way to check this is to view the API in the web browser view and look at the Django Debug Toolbar.  Many simple repeated DB calls can be eliminated with proper use of `select_related` and `prefetch_related`.  Some more advanced cases may require further tuning.

squarelet/organizations/fe_api/serializers.py
allanlasser: Yeah, this is definitely a property we want to have.

squarelet/organizations/fe_api/serializers.py
duckduckgrayduck: @allanlasser What do you think?

squarelet/organizations/fe_api/serializers.py
duckduckgrayduck: I've added users = serializers.PrimaryKeyRelatedField(
        many=True,
        read_only=True,
    )
    for members, but I think I am missing something in relation to admins. 
    
    I don't see a field on the Organization model that lists the admins. Should I add one?

PR #378, opened 2025-08-05
Adds filterable list of subtypes to organization list results

allanlasser:
Third-party clients want to be able to filter the list of organizations by subtype, so that they can better classify verified organizations using their service.

Related to #377 

## Test steps
1. Sign into the preview deployment
2. Go to `/api/organizations`
3. Try filtering by subtype

### DIFF
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -11,7 +11,6 @@ from squarelet.organizations.models import (
     Charge,
     Membership,
     Organization,
-    OrganizationSubtype,
 )
 
 
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -7,12 +7,43 @@ from rest_framework import serializers, status
 from rest_framework.exceptions import APIException
 
 # Squarelet
-from squarelet.organizations.models import Charge, Membership, Organization
+from squarelet.organizations.models import (
+    Charge,
+    Membership,
+    Organization,
+    OrganizationSubtype,
+)
+
+
+def resolve_subtype_names_to_pks(subtype_names):
+    """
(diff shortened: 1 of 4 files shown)


squarelet/organizations/serializers.py
```diff
@@ -7,12 +7,18 @@
...
+    Membership,
+    Organization,
+    OrganizationSubtype,
+)
 
 
 class OrganizationSerializer(serializers.ModelSerializer):
     uuid = serializers.UUIDField(required=False)
     merged = serializers.SlugRelatedField(read_only=True, slug_field="uuid")
+    subtypes = serializers.StringRelatedField(many=True)
```
mitchelljkotler, line 21: This could be a `SlugRelatedField` pointed to name if you want only the name to show up in the results - the `str` method for SubType includes the Type name, which I think looks nice.

squarelet/organizations/viewsets.py
```diff
@@ -15,8 +15,7 @@
 
 
 class OrganizationViewSet(viewsets.ModelViewSet):
-    # remove _plan after clients are updated
-    queryset = Organization.objects.select_related("_plan")
+    queryset = Organization.objects.prefetch_related("subtypes")
```
mitchelljkotler, line 18: `_plan` is no longer needed, but prefetching `subtypes` prevents an additional SQL query for every result returned in the list view

mitchelljkotler (commented):
Using the built in support for related fields reduces code and makes the intentions clearer.

squarelet/organizations/serializers.py
allanlasser: Yeah, I agree having the type name is helpful!

PR #379, opened 2025-08-07
List admins for organizations in API responses

allanlasser:
Clients have requested the ability to see information about the admins for an organization.

The `admins` property on organizations is an object list:
```
[
  {
    "id": user.pk,
    "name": user.name,
    "email": user.email
  }
]
```

### DIFF
--- a/squarelet/organizations/viewsets.py
+++ b/squarelet/organizations/viewsets.py
@@ -15,7 +15,9 @@ from squarelet.organizations.serializers import (
 
 
 class OrganizationViewSet(viewsets.ModelViewSet):
-    queryset = Organization.objects.prefetch_related("subtypes")
+    queryset = Organization.objects.prefetch_related(
+        "subtypes__type", "users__memberships"
+    )
     serializer_class = OrganizationSerializer
     permission_classes = (ScopePermission | IsAdminUser,)
     read_scopes = ("read_organization",)


mitchelljkotler:
> 
> > Looking more deeply, is there a reason why we have parallel APIs?
> 
> As I remember, we have a few reasons for maintaining parallel APIs:
> 
>     1. The "Client API" is read-only, and we don't have plans to add writable endpoints. The frontend API supports POST methods.
> 
>     2. We want to keep the Client API very stable and avoid making any breaking changes. The frontend API only serves our own needs, so we have more flexibility to break and adapt it to meet the needs of our front-end interface.
> 
> 
> I think it continues to make sense to maintain "public" and "private" interfaces, but there's likely serializers and viewsets we could consolidate between both.
> 
> To the point about long-term stability, we should make sure the approach I took here of making an object with admin information is the approach we'd want to maintain long-term.

The main reason is permissions.  The Client API was designed first and foremost for use by our first party clients which we both fully trust and have very different needs from a front end or user facing API.  Trying to combine them into one API seems like a big mess permission wise and likely to result in security issues.  Configuring the permissions for third party clients is more reasonable, as they will have less permissions, but similar needs, and they will be trusted third parties for now.

As far as read only, the Client API does have some write endpoints.  You can create a new user, which powers MuckRock mini-reg system (allowing someone to sign up for a new account from the request composer page).  Client can also create charges, which allows them to sell ala carte options from their site and use the billing information from Accounts.  I imagine the first one is not something we need to give third party clients access to, while the second one is for clients which are using us as their billing provider.

In terms of stability, now that third part clients will be using the client API, it is important to maintain backward compatibility.  The front-end API is just for us for now, but at some point that could become the public API at which point we would want to maintain backward compatibility for it as well.

I think the approach here for exposing admins in the Client API is good and something we can support long term.

mitchelljkotler (approved):

PR #380, opened 2025-08-08
Organization search

eyeseast:
New org search page and component. This introduces Svelte to our stack and outlines a pattern we can use across the codebase.

Files are named consistently, so:
- `squarelet/templates/organizations/organization_list.html` is a Django template
- `frontend/views/organization_list.css` is the CSS file for that template
- `frontend/views/organization_list.ts` is a Typescript entrypoint for that template

Importantly, I'm using Vite to handle all static assets on this page, so `organization_list.ts` imports `organization_list.css` and any other css files it needs, and Vite bundles everything. Then on the template, I use one line to get everything:

```
{% vite_asset 'frontend/views/organization_list.ts' %}
```

Each file in `views` is an entrypoint, so putting a file in `frontend/views` will ensure it gets built and bundled.

### DIFF
--- a/squarelet/templates/organizations/organization_list.html
+++ b/squarelet/templates/organizations/organization_list.html
@@ -74,7 +74,7 @@
         <p>
           {% blocktrans %}
           You must verify your email address before you can send requests
-          or receieve invitations to join an organization.
+          or receive invitations to join an organization.
           {% endblocktrans %}
         </p>
       </div>
@@ -113,7 +113,6 @@
         {% endif %}
 
         {% if pending_invitations %}
-        {{pending_invitations}}
         <div class="pending">
           <header>
             <h2>{% trans "Pending invitations" %}</h2>
--- a/frontend/views/organization_list.css
+++ b/frontend/views/organization_list.css
@@ -17,7 +17,6 @@
   gap: 2rem;
   box-sizing: border-box;
 }
-  
 
 .org-list-page > header {
   width: 100%;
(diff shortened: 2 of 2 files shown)


squarelet/templates/organizations/organization_list.html
```diff
@@ -1,50 +1,96 @@
...
-                  {{ org.name }}
-                </span>
-              </div>
-            </a>
-          {% endfor %}
+      {% endif %}
+
+    {% if invitations %}
+    <div class="pending">
+      <h2>{% trans "Pending requests" %}</h2>
```
allanlasser, line 70: This isn't distinguishing between invitations and requests right now, but each variant of the "invitation" model has distinct actions and relationship to the user.

1. An invitation is sent to the user, and has actions like "Accept" or "Reject"
2. A request is sent from the user, and has a "Withdraw" action.

We should probably be passing two different data through the context for these, and rendering them as distinct sections with distinct actions.

squarelet/templates/organizations/organization_list.html
```diff
@@ -1,50 +1,97 @@
...
+      <h2>
+        {% trans "Pending invitations" %}
+      </h2>
+      {% if not has_pending %}
+      <div class="empty">
+        <p>{% trans "You don't have any outstanding invitations." %}</p>
+        <p>{% trans "Any invitations you receive will appear here." %}</p>
+      </div>
+      {% endif %}
+
```
allanlasser, line 41: While looking at the code for user.get_pending_invitations(), it looks like we return none when the user doesn't have a confirmed email. A check for that here, and a link to send an email confirmation message, would be a great enhancement here.

allanlasser (approved):

PR #399, opened 2025-08-28
Adds API endpoint on staging/dev environment to add redirect URIs

duckduckgrayduck:
This enables MuckRock staging review apps to add their URIs on squarelet staging with an admin account. 

Should be able to go to the web from for one of your local clients and do a POST with a payload like this

{
    "action": "add",
    "redirect_uris": [
        "https://example.com",
    ]
}


duckduckgrayduck:
Should be set for Monday.

mitchelljkotler (approved):

PR #435, opened 2025-09-16
Fix wix sync

mitchelljkotler:
Add tests

Main issue in production was I had set the SITE_ID to the test site instead of the production site


allanlasser (changes requested):

allanlasser (approved):
Looks great!

PR #437, opened 2025-09-16
Fixes visibility of pending invitations

allanlasser:
While working on documentation updates, I realized we weren't displaying invitations appropriately.


allanlasser:
This has come up in a few different contexts, so I've created a new issues to tackle it! https://github.com/MuckRock/squarelet/issues/441

PR #438, opened 2025-09-16
Limit responses of private plans

### LINKED ISSUE #420: Limit responses of private plans
Some of our plans have a boolean `private` flag, along with a `private_organizations` relationship. We don't want private plans appearing to everyone who visits the page.

If the plan is private, we should check if the user is an admin of the `private_organization`. If they are, we should show them the page. If they aren't, we should return a 404.

allanlasser:
Fixes #420

### DIFF
--- a/squarelet/payments/views.py
+++ b/squarelet/payments/views.py
@@ -19,9 +19,32 @@ from squarelet.organizations.models import Organization, Plan
 logger = logging.getLogger(__name__)
 
 
+def protect_private_plan(plan, user):
+    """Raise 404 if user should not access this private plan"""
+    if plan.private_organizations.exists():
+        # If user is not authenticated, raise 404
+        if not user.is_authenticated:
+            raise Http404("Plan not found")
+
+        # If user is not admin of any organization for this plan, raise 404
+        user_admin_orgs = user.organizations.filter(memberships__admin=True)
+        if not plan.private_organizations.filter(pk__in=user_admin_orgs).exists():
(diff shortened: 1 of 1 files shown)


squarelet/payments/views.py
```diff
@@ -19,10 +19,33 @@
...
+            queryset = self.get_queryset()
+
+        plan = super().get_object(queryset)
+
+        protect_private_plan(plan, self.request.user)
+        return plan
+
     def get_context_data(self, **kwargs):
         # pylint: disable=too-many-locals
         context = super().get_context_data(**kwargs)
```
mitchelljkotler, line 51: Do we need to limit `admin_orgs` to only orgs the user is an admin in which are also in `private_organizations` if `private_organizations` is set?  Otherwise if we whitelist a plan for one organization, an admin for that organization could subscribe another organization they are an admin of for that plan.

mitchelljkotler (commented):
This looks good, but I want to make sure I understand how it works:

If a plan is private and has private organizations set, only users in those organizations will be able to see the plans from either their organization upgrade page or the plan's dedicated page.  (This is our existing private plans).

If a plan is private but does not have private organizations set, no one will see it on their organization upgrade page, but anybody can view it's dedicated page.  (This is the new Sunlight plans).

If a plan is not private, everyone can view it in all places.  (Professional and Organization plans).

allanlasser:
The review app successfully built and deployed, so we can test our assumptions hold before merging.

PR #439, opened 2025-09-17
Limits visible organizations

### LINKED ISSUE #322: Limit the display of public organizations
From @morisy:

> I don’t think this spilled over to Accounts, but on MuckRock profiles were private until you filed at least one request because the one challenge is spam. If you have some place people can make anything public, people will try to spam it (the projects tool is filled with spam which I was not expecting).
> 
> [When creating new organizations,] the default should be public, but not necessarily actually making it publicly visible on Accounts until:
>
> - They’re verified, or
> - They’ve filed a request.

We should add a new derived property to the `Organization` model, `is_visible`, that applies additional criteria on top of the boolean `private` property.

allanlasser:
Fixes #322

- Adds conditions to the `OrganizationQuerySet.get_viewable` method to only return organizations that have been verified or have exisiting charges.
- Adds test coverage for `OrganizationQuerySet`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,8 +163,8 @@ typings/
 !.vscode/launch.json
 !.vscode/extensions.json
 
-
-
+### Claude Code
+CLAUDE.md
 
 
 ### Windows template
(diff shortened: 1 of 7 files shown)


duckduckgrayduck (approved):
This functions as described.

PR #462, opened 2025-10-03
Invoicing enhancements

### LINKED ISSUE #415: Support purchasing annual plans with immediate payment
When someone purchases an annual plan, we use an invoicing payment flow, even if we have captured a card for them or have one on file.

https://github.com/MuckRock/squarelet/blob/444b6ce94b384755bc1b9e59f11da72685f77f47/squarelet/organizations/models/payment.py#L204-L206

This has a few issues:

1. We require credit card information when subscribing to an annual plan, but we never charge it.
2. Often, people want to make an immediate purchase. Business customers are incentivized to make large purchases with credit cards for points, security, or other kickback benefits.
3. We create additional work for ourselves by needing to follow up on issued invoices.
4. We have no way of checking if an invoice has been paid, or restricting functionality when an account is overdue.

If somebody is purchasing an annual plan, there's no reason we can't charge a credit card for the purchase as an alternative to an issued invoice.

When purchasing an annual plan, we may want to provide `Send me an invoice` as a valid payment method, in addition to paying with a card on file or paying with a new card.

### LINKED ISSUE #461: Handling for overdue invoices
We originally flagged in #415 that while we have existing logic for handling overdue/failed payments for credit cards, we don't have any logic for when an invoice becomes overdue. We want to extend our existing overdue logic to apply to unpaid invoices, as well.

allanlasser:
This update resolves two issues with how we handle invoices as a payment method:

- Resolves #415 by supporting credit card payments for annual plan. It provides a new payment choice in the Plan payment form, "Pay by invoice." When explicitly selected, an invoice will be issued to the customer for payment. It also uses invoices as a fallback method when a card token is missing.
- Resolves #461 with lifecycle tracking and automated handling for overdue invoices. This brings invoice-based payments for annual plans to parity with credit card payment monitoring when it comes to overdue payment or cancellation. It also adds some useful features for admins to observe and interact with invoices over their lifecycle.

**For a detailed list of changes and instructions for manual testing, see our [release plan in Notion](https://www.notion.so/muckrock/Invoicing-Enhancements-281f889269638001849feb13cba35051).**

This is a complex PR with an external dependency on Stripe webhooks, so it may take some time to thoroughly review this.


allanlasser:
@duckduckgrayduck caught that if we fail to receive the webhook, we'll never have an internal record of the invoice. 

I’ll look into creating a stub Invoice object when the subscription starts; maybe the Stripe call returns an ID we can save. If the Stripe call fails, we’ll definitely need to catch that in Sentry and probably not allow the transaction to succeed.

squarelet/organizations/admin.py
```diff
@@ -282,6 +366,28 @@ def get_email_domains(self, obj):
...
+        count = getattr(obj, "outstanding_invoice_count", 0)
+        total = getattr(obj, "outstanding_invoice_total", 0)
+        if count > 0:
+            return f"{count} (${total / 100:.2f})"
+        return "-"
+
+    get_outstanding_invoices.short_description = "Outstanding Invoices"
+    get_outstanding_invoices.admin_order_field = "outstanding_invoice_count"
+
+    def is_visible_to_anonymous(self, obj):
```
mitchelljkotler: This is adding a query per organization the the list display page and may affect performance

allanlasser: I'm fine removing this column, since we have another admin view for seeing invoices.

PR #479, opened 2025-10-21
When staff take the "leave org" action on a user account, remove the user not staff member

### LINKED ISSUE #476: When staff take the "leave org" action on a user account, it removes the staff user from the org
When a staff tries to remove a user from an org using the "Leave" action from their account page, the handler assumes the signed in user is the one the to remove from the org. Instead, we should either:
- include a user ID as a hidden input in the form and use that to remove the user from the org.
- make sure the handler is applied to the page for the current user, not the signed in user.

allanlasser:
Fixes #476

### DIFF
--- a/squarelet/organizations/views.py
+++ b/squarelet/organizations/views.py
@@ -108,13 +108,10 @@ class Detail(AdminLinkMixin, DetailView):
 
         return context
 
-    def post(self, request, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


duckduckgrayduck (approved):
When I login on staging as a staff account, hijack a normal user, and leave the org, it correctly has the normal user leave the org, and not the staff account

PR #501, opened 2025-11-05
Admin event logging

duckduckgrayduck:
Copies new_action from muckrock/utils. 

Log


allanlasser (commented):
Just chiming in to say that this looks great, and I'm excited to have this as a foundation for exposing site activities to other users!

PR #506, opened 2025-11-17
Update organization profiles with staff approval

### LINKED ISSUE #468: Support editing organization profiles with staff approvals
We have a page for admins to edit the properties of their organization’s profile. This information will be displayed on the organization’s profile, if it’s public.

[This mockup shows how the form will appear to an org admin.](https://www.figma.com/design/9KXyTVD6KfAHnXvWSbapv0/Accounts?node-id=2797-11488&m=dev)

This mockup shows how the form will appear to an org admin.

Org admins may freely edit these organization fields:

- `avatar`, using our `AvatarWidget`
- `bio` with information about for the organization.
- `private` checkbox:
    - Private is false by default.
    - If the organization is public and in a hidden state (following proposal in [[User and organization privacy](https://www.notion.so/User-and-organization-privacy-279f889269638051ac3ece39faf69e7d?pvs=21)](https://www.notion.so/User-and-organization-privacy-279f889269638051ac3ece39faf69e7d?pvs=21)), show a message explaining that they will not appear publicly until the org is verified or has made a payment.

Org admins should be able to request changes to the following data:

- `name`
- The `slug` used to identify the organization in URLs.
    - Like with individual profiles, slugs may be changed once before editing them is disabled. If an org admin would like to change their slug again, they need to contact support with a change request.
    - When we can’t find an org matching a slug, we can query the table of approved change requests to find an older slug that matches.
    - Since changing a slug will break existing URLs to the organization’s pages in Accounts, we will handle broken links by showing signed-in users a list of their orgs on any organization 404 pages.
- `url`
- Location

When requesting a change, we’ll also want to [prompt](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt) users for an explanation of their changes. After the user clicks the “Request changes” button, we can raise a [`prompt`](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt) to captures this explanation before continuing with the submission.

When a staff member saves the form, we don’t need an explanation and don’t need to issue a change request. For staff members, the button should read “Save Changes”.

## Tracking change requests

We require staff confirmation for names and URLs to prevent impersonations. When changes are requested, we will create a new support ticket with a link to the organization’s profile.

Each change to an organization’s profile data should be recorded in a new table, `OrganizationProfileChangeRequests`*.* The model should include fields for:

- a unique `id`
- the `user` who made the request
- a relation to the `organization`
- an `explanation` for the change
- a `status` for the change (one of: `pending`, `accepted`, `rejected`)
- “previous” data, pulled from the organization during creation:
    - `old_display_name`
    - `old_slug`
    - `old_url`
    - `old_location`
- the “changed” data, possibly null if not different from previous data:
    - `new_display_name`
    - `new_slug`
    - `new_url`
    - `new_location`

When a change request is created, we will also create a support ticket with [Zenpy](https://github.com/facetoe/zenpy) and the ZenDesk API. [See the `muckrock` repo for a reference implementation](https://github.com/MuckRock/muckrock/blob/9e9ece3017996f85b898535e088a4a1212544647/muckrock/task/models.py#L790-L888) using the [Zenpy](https://github.com/facetoe/zenpy) library. This will allow staff to respond to the user to discuss their request.

The support request can look something like:

```
<user.name> requested the following changes to their organization <organization.name>:

- Display Name: New Yor Times -> New York Times
- Slug: new-yor-times -> new-york-times

They provided the following explanation:

> We had a typo in our org name.

To approve this change, please follow this link:

https://accounts.muckrock.com/organization/<slug>/change-request/<id>
```

Staff will have permission to make the edit on this page, so they can make any adjustments they need to before “Confirming” the changes by saving the form.

When the form is saved, the changed data should be applied to the Organization.

eyeseast:
Closes #468 

**Migrations required**

This adds a new model -- `organizations.ProfileChangeRequest` -- where org admins can request changes to their core profile, plus a form and views to process those proposed changes.

To do:

- [x] restyle forms to match new designs
- [x] create a ZenDesk ticket pointing back to the org update page


squarelet/organizations/forms.py
```diff
@@ -259,3 +271,91 @@ def clean(self):
...
+                f"Current: {org.get_country_display() or 'Not set'}. "
+                "Leave blank to keep unchanged."
+            )
+            self.fields["url"].help_text = _(
+                "Add a URL to associate with this organization."
+            )
+            self.fields["explanation"].help_text = _(
+                "Explain why you are requesting these changes "
+                "(required for staff review)."
+            )
```
allanlasser: As a user, it feels clearer to me to have the fields prefilled with their current value instead of with the placeholder + help text approach. We could then check for value changes when cleaning the field.

allanlasser (changes requested):
Overall this looks really good! There's a few things I caught during the review. I also merged recent updates to `master` into `redesign-orgs` and this branch, cleaning up a few merge conflicts along the way.

I like the direction you went with the Crispy Form components, but it looked to me like just `input` has been modified from its original version so far—correct? I think a README in the `templates/forms` directory would go a long way to explaining to future us why we went this direction!

squarelet/templates/organizations/organization_form.html
```diff
@@ -11,6 +9,136 @@
...
+          </button>
+        </form>
+      </div>
+    </section>
+  </div>
+
+</article>
+{% endblock content %}
+
+{% comment %}
```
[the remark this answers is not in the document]
eyeseast: Yeah, these should be different templates if they're different views. And we probably want to rename them to match how we're organizing files now.

squarelet/organizations/forms.py
```diff
@@ -259,3 +271,91 @@ def clean(self):
...
+
+
+class ProfileChangeRequestForm(forms.ModelForm):
+    """Request changes to core organization profile data"""
+
+    url = forms.URLField(label=_("URL"), required=False)
+
+    class Meta:
+        model = ProfileChangeRequest
+        fields = ["name", "slug", "url", "city", "state", "country", "explanation"]
```
allanlasser: Explanation should be a required field for non staff users when the form renders. This way, the browser can enforce the requirement before submission.

allanlasser (approved):
I removed the `{% comment %}` block from the `organization_update.html` template—everything else looks good to me.

PR #513, opened 2025-12-02
Track rejected requests to join and present them to the user

### LINKED ISSUE #484: Prevent users from re-requesting to join an organization when they have a pending or declined request
We know when a user has a pending request: it's an invitation where they're they user, the "request" flag is up, and there's no accepted or rejected date.

We also know when a user has a rejected request: it's an invitation where they're they user, the "request" flag is up, and there's a rejected date.

When a user has a pending or rejected request, we should show them a message indicating this and not show a "Request to join" button.

duckduckgrayduck:
When a user has sent a request to join they are already blocked from sending another one via the requested_invite logic. This adds a similar blocker if the user has a rejected request to join an organization and lets them know to contact an admin if this is a mistake and they do indeed need access. Should close #484


squarelet/organizations/querysets.py
```diff
@@ -150,6 +147,9 @@ def get_pending_invitations(self):
     def get_pending_requests(self):
         return self.get_open().filter(request=True)
 
+    def get_rejected_requests(self):
```
allanlasser: We don't have any tests for the InvitationQuerySet now—it'd be worthwhile to build coverage while we're touching this, but I'll leave that up to you.

allanlasser (commented):
Looks great, nice and simple fix that we don't have to wait to ship. Two small pieces of feedback here.

squarelet/organizations/querysets.py
duckduckgrayduck: Good call- I'll write some tests!

PR #514, opened 2025-12-03
Rate limit join requests

duckduckgrayduck:
RateLimitMessage

### DIFF
--- a/squarelet/organizations/views.py
+++ b/squarelet/organizations/views.py
@@ -149,8 +149,8 @@ class Detail(AdminLinkMixin, DetailView):
             zendesk_description=(
                 "The following user has reached the "
                 "rate-limit for joining organizations, "
-                "sending {recent_requests} requests in the last "
-                f"{settings.ORG_JOIN_REQUEST_WINDOW} seconds:\n\n"
+                f"sending {settings.ORG_JOIN_REQUEST_LIMIT} requests in the last "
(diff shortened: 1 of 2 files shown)


squarelet/organizations/views.py
```diff
@@ -119,37 +122,74 @@ def get_context_data(self, **kwargs):
...
-                    ),
+        user = request.user
+
+        # Already a member
+        if self.organization.has_member(user):
+            return
+
+        # Auto join if allowed
+        if user.can_auto_join(self.organization):
+            self.organization.memberships.create(user=user)
```
allanlasser: Seeing how this is simplified here, I'm wondering if the sync to Wix would work better as a signal on the membership creation—this way, creating memberships through other means would also trigger a sync for our Sunlight member orgs. Thoughts?

squarelet/organizations/views.py
```diff
@@ -119,37 +122,75 @@ def get_context_data(self, **kwargs):
...
+                sync_wix.delay(
+                    self.organization.pk,
+                    self.organization.plan.pk,
+                    user.pk,
                 )
-                invitation.send()
+            messages.success(request, "You have successfully joined the organization!")
+            return
+
+        # For normal requests to join, check rate limit
```
mitchelljkotler: This method is getting a little long, I think factoring out the rate limiting logic into its own method would be helpful.  Would also make it easier to reuse in the future if we needed to rate limit other actions.

mitchelljkotler (changes requested):

PR #525, opened 2025-12-17
Expose org management links to MuckRock staff

allanlasser:
As a staff user, I can view the org editing and member management pages, but I can't find them from the frontend.

### DIFF
--- a/squarelet/organizations/tests/test_views.py
+++ b/squarelet/organizations/tests/test_views.py
@@ -86,6 +86,23 @@ class TestDetail(ViewTestMixin):
         assert isinstance(response.context_data["invite_count"], int)
         assert response.context_data["invite_count"] == 3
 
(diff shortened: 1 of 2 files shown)


duckduckgrayduck:
When I search for an organization to test this on and click on the organization, I receive a 500 error. https://muckrock.sentry.io/issues/7122148094/?project=1247105&query=is%3Aunresolved&referrer=issue-stream

duckduckgrayduck:
Since this is a staff action, we should also log it similar to hijacks and editing a member page as a staff user. This is new_action in core/utils.py 
https://github.com/MuckRock/squarelet/commit/b635350adf459a9775d1ac83beabebc462e36d96#diff-c4d896b0e329737bfc2878b9c57885e6501d5d48902766be6c0a3f5e30100e60

PR #534, opened 2026-01-09
Show users a history of their org invites and join requests

### LINKED ISSUE #486: Show users a history of their join requests
We’d like to add a screen that lets users see a history of all of their requests and invitations. This is a pretty simple database query: Invitations where user = X, split into “requests” and “invitations”. We’d make this accessible with a link from the “Organizations” page where we show users an overview of all their memberships and pending requests/invitations.

I started sketching this out in Figma, but we’ll be able to more rapidly prototype this in code. We could either render this as a list of `TeamListItems` with additional annotations, or keep it simpler and render a table with a historical record of all invitations/requests, their status, and their date—we can try both approaches in code and compare them to see which is more legible.

If a user previously declined an invitation, they can send a new request to join the organization.

allanlasser:
Closes #486

Gives users a deep, paginated history of their invitation and join requests

### DIFF
--- a/frontend/css/user_invitations.css
+++ b/frontend/css/user_invitations.css
@@ -8,7 +8,19 @@
   min-height: 100%;
 }
 
+.invitation-item, .invitation-item main, .invitation-details {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 0.5rem 1rem;
+}
+
+.invitation-item header {
+  flex: 1 1 16rem;
+}
+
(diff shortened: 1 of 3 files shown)


allanlasser:
> The request and invitation pages look great. I don't see a way to get to them, though (unless you know the URL).

Right now there are links to these pages on `/organizations`, in the sections where we show your pending invitiations/requests (or empty state if none).

Where else would you expect to find links to these pages?

PR #546, opened 2026-01-16
Supports non-profit Sunlight memberships

### LINKED ISSUE #538: Discount plans for non-profits
We want to be able to sell Sunlight memberships to non-profit newsrooms at a discount. We are OK with newsrooms self-reporting their non-profit status and not requiring staff to gatekeep this discount. When purchasing a plan, newsrooms should be able to check an option for "We are a non-profit" and receive a discounted subscription to an annual or monthly plan.

allanlasser:
Resolves #538

- Introduces new non-profit variant plans which users may opt-in to when purchasing a Sunlight membership.
- Migrates existing Sunlight plans to use new copy and creates new non-profit variant plans locally and in Stripe.
- Updates the `/selectplan/` page to reflect changes to Sunligiht plan structure.
- Updates handling for Enterprise plans to refer interested users to scheduling a meeting with us.


squarelet/organizations/migrations/0050_update_sunlight_plans_and_add_nonprofit_variants.py
```diff
@@ -0,0 +1,305 @@
...
+            "new_slug": "sunlight-enterprise-annual",
+            "name": "Sunlight Research Center - Enterprise (Annual)",
+            "base_price": 12000,
+            "short_description": enterprise_short_description,
+            "benefits": enterprise_benefits,
+        },
+    ]
+
+    for update in plan_updates:
+        Plan.objects.filter(slug=update["old_slug"]).update(
```
mitchelljkotler: Updating the slug of existing plans will break their stripe ID.  We would either need to update the stripe ID or not change existing slugs

squarelet/payments/views.py
```diff
@@ -207,6 +242,36 @@ def post(self, request, *args, **kwargs):
         new_organization_name = request.POST.get("new_organization_name")
         stripe_token = request.POST.get("stripe_token")
         payment_method = request.POST.get("payment_method")
+        is_nonprofit = request.POST.get("is_nonprofit") == "on"
```
mitchelljkotler: This view really needs to use a Django form for validation.  We can release this as is for now for time reasons, but should not continue to work on this view without refactoring.

PR #570, opened 2026-02-03
Create new PlanPurchaseForm with resuable templates

### LINKED ISSUE #522: Refactor a PlanPaymentForm from `payment/views.py`
Right now we use naive POST handling for the payment form. Refactoring this logic into a PlanPaymentForm will make our code simpler and allow us to reuse this form in more places.

We should also makes sure that this refactored form incorporates the template logic and JS logic in `templates/payments/plan.html`, since it's tightly coupled to the form behavior.

allanlasser:
Fixes #522

Refactors our `plan.html` and `payments/views.py` logic into a `PlanPurchaseForm` that renders a `plan_purchase.html` template. This isolates our payment handling logic, makes it more reusable, and is more aligned with Django's approach.

### DIFF
--- a/squarelet/payments/tests/test_forms.py
+++ b/squarelet/payments/tests/test_forms.py
@@ -259,30 +259,40 @@ class TestPlanPurchaseFormValidation:
         assert form.is_valid(), f"Form errors: {form.errors}"
 
 
-@pytest.mark.django_db
-class TestPlanPurchaseFormTemplateErrors:
-    """Test that form errors are rendered in the template"""
+class TestPlanPurchaseFormTemplate:
+    """Test that the template renders as expected"""
+
+    TEMPLATE_PATH = (
+        Path(__file__).resolve().parents[2]
+        / "templates"
+        / "payments"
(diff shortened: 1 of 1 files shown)


squarelet/payments/forms.py
```diff
@@ -0,0 +1,362 @@
...
+                    nonprofit_plan = Plan.objects.get(slug=nonprofit_slug)
+                    data["nonprofit_base_price"] = nonprofit_plan.base_price
+                    data["nonprofit_price_per_user"] = nonprofit_plan.price_per_user
+                    data["has_nonprofit_variant"] = True
+                except Plan.DoesNotExist:
+                    data["has_nonprofit_variant"] = False
+            else:
+                data["has_nonprofit_variant"] = False
+        else:
+            data["has_nonprofit_variant"] = False
```
mitchelljkotler: I think setting `data["has_nonprofit_variant"] = False` as the default, and overriding it to True when true and then leaving off the else clauses would make this slightly cleaner

PR #585, opened 2026-02-11
Redirect user correctly after accepting invite

### LINKED ISSUE #576: Accepting an invite gives us an alert, "Invitation accepted" but the page is shown as not found
Should probably redirect back to the user's profile or the org's landing page

duckduckgrayduck:
Closes #576

### DIFF
--- a/squarelet/organizations/views/members.py
+++ b/squarelet/organizations/views/members.py
@@ -271,10 +271,15 @@ class InvitationAccept(DetailView):
         invitation = self.get_object()
         action = request.POST.get("action")
         if action == "accept":
-            org = invitation.organization
             invitation.accept(request.user)
             messages.success(request, "Invitation accepted")
-            return redirect("organizations:detail", slug=org.slug)
+
+            # If the referer is this invitation page itself, ignore it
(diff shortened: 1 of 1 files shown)


squarelet/organizations/views/members.py
```diff
@@ -271,9 +271,10 @@ def post(self, request, *args, **kwargs):
         invitation = self.get_object()
         action = request.POST.get("action")
         if action == "accept":
+            org = invitation.organization
             invitation.accept(request.user)
             messages.success(request, "Invitation accepted")
-            return get_redirect_url(request, redirect(invitation.organization))
```
allanlasser: I remember adding this utility because you can accept an invitation from different places, and this is intended to keep you on the page you were on, with a fallback destination if there's no referrer in the HTTP header.

We should keep this the way it is, since when you accept an invitation from your account page, the organization list page, the organization detail page, or the onboarding flow, you expect to stay on that page. The only page we _don't_ want to stay on is the invitation page.

The easiest way to do that would probably be to remove the HTTP_REFERRER header if it matches our invitation's route. We'll also want to make sure that behavior is consistent for the `reject` action, too.

PR #587, opened 2026-02-12
Resize images wherever we can

### LINKED ISSUE #469: Check sizes of images we're serving
We have `sorl.thumbnail` installed and we load the `{% thumbnail %}` template tag on lots of pages, but as far as I can tell, we're not actually using it. Let's check that we're not accidentally serving giant images as avatars.

eyeseast:
Closes #469 

Wherever possible, images are now wrapped in `sorl.thumbnail` resizing, either in templates or in Python code. 

Testing Recommendations

  - Verify navigation avatar displays correctly at 25px
  - Check organization dropdown shows 40x40 avatars
  - Test profile edit page shows resized preview
  - Confirm API returns avatar_small and avatar_medium fields
  - Verify graceful fallback when thumbnail generation fails
  - Test with missing/corrupted avatar files

Claude did all of this.

### DIFF
--- a/squarelet/templates/templatetags/services_list.html
+++ b/squarelet/templates/templatetags/services_list.html
@@ -1,8 +1,15 @@
+{% load thumbnail %}
 <div class="service-providers">
   {% for provider in service_providers %}
     {% if provider.base_url %}
       <a href="{{provider.base_url}}" target="_blank" class="service-provider">
-        <img src="{{ provider.icon.url }}" alt="{{ provider.name }}">
+        {% if provider.icon %}
+          {% thumbnail provider.icon "60x60" crop="center" as im %}
+            <img src="{{ im.url }}" width="{{ im.width }}" height="{{ im.height }}" alt="{{ provider.name }}">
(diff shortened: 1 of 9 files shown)


squarelet/templates/templatetags/services_list.html
```diff
@@ -1,16 +1,29 @@
...
   {% for provider in service_providers %}
     {% if provider.base_url %}
       <a href="{{provider.base_url}}" target="_blank" class="service-provider">
-        <img src="{{ provider.icon.url }}" alt="{{ provider.name }}">
+        {% if provider.icon %}
+          {% thumbnail provider.icon "60x60" crop="center" as im %}
+            <img src="{{ im.url }}" width="{{ im.width }}" height="{{ im.height }}" alt="{{ provider.name }}">
+          {% endthumbnail %}
+        {% else %}
+          <img src="{{ provider.icon.url }}" alt="{{ provider.name }}">
```
allanlasser: This looks like it'll throw an error or fail if `provider.icon` is false

PR #588, opened 2026-02-12
Makes withdrawing a request semantically distinct operation

### LINKED ISSUE #575: Withdrawing a request to join should delete it, not reject it
If I withdraw a request to join an organization, I cannot re-request to join it. That's because the invitation is marked as rejected by the user who sent it; it'd be more accurate to delete the request.

allanlasser:
Fixes #575

Previously, we conflated withdrawing a request with rejecting, which led to some unexpected behavior. This is now clearer, with a specific ability to understand what happened to a particular request.

Note: currently based on #582 to avoid migration merge conflicts. Should be rebased and merged following that change.


allanlasser:
Great, fixed that one too—two missing lines in our User model.

PR #616, opened 2026-03-10
User search on manage members

### LINKED ISSUE #474: Search existing users when creating org membership invitations
When inviting members to join an organization, we'd like to enable org admins to search  existing users within our Accounts system. This will require making some changes to user privacy controls and a new Svelect component for interactively searching users.

## User privacy rules

1. There should be three privacy states for individuals: public, private, and hidden.
2. New individuals should be “public” by default when created, but “hidden” until they are either:
    a. verified (either directly or by association)
    b. have made a payment
    c. or have confirmed their email address
3. Individuals may opt-in to their account being “private” on the profile editing page. While “private”, individuals are only visible to MuckRock staff and members of the organization. This includes profile page visibility, visibility in lists and search results.
4. While “public”, individuals can be found through user search as long as they are not “hidden” and have a confirmed email address. Since inviting users requires sending an email, we want to ensure we’re sending to valid emails.
5. While “hidden”, individuals will not be included in search results.

## Make the user's privacy editable

The "private" boolean field is already on individual organizations, but until now it's not been exposed to users. We'll add a new checkbox field to the form for editing the user's individual organization:

[Image](https://www.figma.com/design/9KXyTVD6KfAHnXvWSbapv0/Accounts?node-id=1206-5104&m=dev)

## Selecting users

We'll add a new Svelte components, `UserSelect`, that builds upon the `OrgSearch` component we added previously:

[Image](https://www.figma.com/design/9KXyTVD6KfAHnXvWSbapv0/Accounts?node-id=3096-11322&m=dev)

It has three states:
1. Empty
2. Searching; it will show results as a dropdown list of small [`UserListItem`s](https://www.figma.com/design/9KXyTVD6KfAHnXvWSbapv0/Accounts?node-id=3096-11386&m=dev). If it detects the query has the structure of an email address, and that address doesn't belong to one of the user results, an option to select the email should appear in the dropdown.
3. Selected; it will show selections as chips, styled based on whether a user or an email was selected.

### User invitation form

We should use the `UserSelect` component as a widget in a UserInvitationForm, which will disable the "Invite Users" button until the widget has contents. Upon submission, the form should create invitations for the users and emails selected.

eyeseast:
Closes #474 

Screenshot 2026-04-24 at 4 14 20 PM
Screenshot 2026-04-24 at 4 14 04 PM

Screenshot 2026-04-24 at 4 14 20 PM

## Summary
- Add three-state user privacy (public / private / hidden) with signals to
  auto-unhide users on email verification or payment
- Expose privacy toggle in user profile settings
- Add user search API endpoint (`GET /fe_api/users/?search=`) with visibility filtering
- Build `UserSelect` Svelte component (Svelecte-based, supports both user search
  and raw email entry) and wire it into the org invite-members form
- Progressive enhancement: form falls back to plain email input without JS
- Invitations with existing accounts show user info, not email in invites list

## Test plan
- [x] Verify hidden users don't appear in search results for non-staff
- [x] Verify private users appear only to org-mates
- [x] Verify staff sees all users
- [x] Test email verification signal flips `hidden=False`
- [ ] Test `UserSelect` component: search, select user, type email, remove chip
- [ ] Test invitation POST sends email for both user-type and email-type selections
- [x] Test no-JS fallback still works via Django form POST


squarelet/organizations/fe_api/viewsets.py
```diff
@@ -66,6 +66,38 @@ class InvitationViewSet(
...
+        if not email and user_obj:
+            email = user_obj.email
+
+        if email and org and org.has_member_by_email(email):
+            return Response(
+                {"status": "already_member", "email": email},
+                status=status.HTTP_200_OK,
+            )
+
+        self.perform_create(serializer)
```
allanlasser: This feels like unnecessary encapsulation, since I only see `perform_create` being called once.

allanlasser (commented):
Dropping some notes on backend before moving into frontend review. There's some critical decisions in migration 0060 we should discuss.

allanlasser (changes requested):
Ok, done with frontend review. Contributing a commit with some style tweaks on the input. Going to flag this as changes requested due to comments on backend files.

allanlasser (approved):

squarelet/users/fe_api/viewsets.py
```diff
@@ -1,22 +1,52 @@
...
+    UserSearchSerializerNoEmail,
+    UserSerializer,
+)
 from squarelet.users.models import User
 
 
 class UserViewSet(viewsets.ReadOnlyModelViewSet):
-    serializer_class = UserSerializer
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (AllowAny,)
```
duckduckgrayduck: Is there a reason we are changing to from IsAuthenticated to AllowAny here?

squarelet/organizations/fe_api/serializers.py
```diff
@@ -83,9 +83,8 @@ class Meta:
             "rejected_at",
             "withdrawn_at",
             "created_at",
-            "user",
         )
-
-    def perform_create(self, serializer):
```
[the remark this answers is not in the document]
eyeseast: It was only used once, so I inlined it.

squarelet/users/fe_api/viewsets.py
eyeseast: It was causing a redirect loop when I checked if the user was verified, so now we check later in the process and return an empty queryset for anonymous users.

allanlasser (approved):

PR #621, opened 2026-03-19
Adds local HTTPS config for Minio

allanlasser:
While getting my local DocumentCloud stack working, I ran into an upload error with Minio related to calling an unsecured `http` URL from the `https://www.api.documentcloud.org` server. I updated my local app to use `https://` as the minio address, and local uploads worked with our local SSL config.

This should only impact local development.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -38,6 +38,7 @@ jobs:
         run: |
           echo "127.0.0.1 dev.squarelet.com" | sudo tee -a /etc/hosts
           echo "127.0.0.1 dev.mailhog.com" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 minio.documentcloud.org" | sudo tee -a /etc/hosts
 
       - name: Generate environment files
         run: python initialize_dotenvs.py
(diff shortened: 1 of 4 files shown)


duckduckgrayduck:
Why are we adding minio to squarelet's networks and not documentclouds? the minio container is spun up on DocumentCloud. 

At this moment, I cannot approve this PR as I cannot get this running locally.

duckduckgrayduck:
I got this working with some changes in local.yml and .envs/.local/.django in the DocumentCloud repo. 

- Added squarelet_default network with internal.minio.documentcloud.org alias to documentcloud_minio

- Changed documentcloud_minio's default network alias from minio.documentcloud.org to internal.minio.documentcloud.org

In .envs/.local/.django:

- Changed MINIO_URL from https://minio.documentcloud.org:9000 to https://minio.documentcloud.org

This works 100% on the Django 4 version. It is broken in Django 5 and let me see why this might be

duckduckgrayduck:
I got this working in Django 5, but it requires adding a few more CORS headers in the nginx conf so uploads work for both private and public. It also required some fixes in the DocumentCloud local processing pipeline which I'll include in the django5_upgrade branch and merge into staging. I'll probably need to adapt the environment variable creation to fix this as well

duckduckgrayduck:
I got this working fully with the Django 5 upgrade branch. Required updating the Minio image to something more modern (2024) for SSL to work with the upgraded S3 packages. I added two things in this PR, which we will merge when the Django 5 upgrades are ready to merge. One is upping the client max body size, because nginx limits us to small documents otherwise, and the other is a preference thing to keep minio link the same as what it was before but means I don't have to change it on DocumentCloud's side

duckduckgrayduck (approved):

PR #628, opened 2026-03-24
Support redirecting after successful purchase

### LINKED ISSUE #312: Our allowed hosts check won't update as we add new clients
Ran into this while working with our `next` URL forwarding. I expect we will get bitten by needing to merge code any time we add a new client app's domain. Is there a more dynamic way we can generate this list, like from a database query?

https://github.com/MuckRock/squarelet/blob/63217a7e26dcf85d5f6896aae03d212c1a2b7464/squarelet/users/adapters.py#L51-L62

### LINKED ISSUE #623: Support post-purchase redirects for plans
If a user is purchasing a plan for a third-party service, we want to give them a way to get back to the place where they triggered the purchase from. We can add a `?purchase_redirect=URL` query parameter that sends users to the provided URL after the purchase succeeds.

allanlasser:
Fixes #623

Adds a `purchase_redirect` query param that is validated/cleaned and added to the purchase form. When the purchase succeeds, the user is redirected to this URL instead of their organization page.

Fixes #312

Updates our definition of safe URLs to include Service base URLs and OIDC client redirect URLs.

### DIFF
--- a/squarelet/payments/forms.py
+++ b/squarelet/payments/forms.py
@@ -10,8 +10,8 @@ import logging
 import sys
 
 # Third Party
-from allauth.account.adapter import get_adapter
 import stripe
+from allauth.account.adapter import get_adapter
 
 # Squarelet
 from squarelet.core.forms import StripeForm


squarelet/payments/forms.py
```diff
@@ -245,6 +268,16 @@ def clean_new_organization_name(self):
 
         return name
 
+    def clean_purchase_redirect(self):
```
mitchelljkotler, line 271: I think we want to restrict this more to known client sites, otherwise we open it up to phishing attacks.  There is a `is_safe` method in `users/adapters.py` which checks for this, but that method probably needs to be updated to check all `Service` model URLs rather than being hard coded (eventually).

allanlasser: We already have an issue for this (#312) and we might as well fix this now.

PR #630, opened 2026-03-27
Create/edit subscriptions from Organization admin

allanlasser:
Updates the inline subscriptions in the org admin to support creating new subscriptions and editing existing ones. This will make it easier to associate custom enterprise plans with Sunlight parent orgs.

### DIFF
--- a/squarelet/organizations/admin.py
+++ b/squarelet/organizations/admin.py
@@ -14,10 +14,9 @@ import logging
 import sys
 from datetime import date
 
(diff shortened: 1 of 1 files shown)


allanlasser:
My goal here was to _circumvent_ payment, and create a subscription to a plan for an enterprise customer, with an assumption that payment would be handled through a backchannel. But in this case, I guess we just create a private plan with a price of `0` and then hijack for enrollment. Overall I think you're right here, will close.

PR #641, opened 2026-04-13
Remove Wix labels on membership/sub changes

### LINKED ISSUE #589: Remove Wix labels for membership and subscription changes
## Problem

The Wix integration is currently one-way: users are synced to Wix (labels added, members created) when they join a Wix plan organization, but there is no cleanup when they leave. This means users retain stale Wix labels like `custom.paying-member` and tier-specific labels (e.g., `custom.essential-member`) after they should have lost access.

### Affected scenarios

1. **Membership removal** — When a user is removed from an organization on a Wix plan, their Wix labels are not removed. There is an existing TODO in `squarelet/organizations/models/membership.py` acknowledging this:
   ```python
   # TODO: We need to remove somebody's Wix membership here too
   ```

2. **Subscription cancellation** — When an organization's subscription is cancelled (e.g., due to payment failure via `subscription_cancelled`), the Stripe subscription is cleaned up but Wix labels for all org members remain.

3. **Plan downgrade** — When an organization switches from a higher tier to a lower tier, new labels are added but old tier labels are not removed.

## Expected behavior

- When a user is removed from a Wix plan organization, their Wix labels should be removed.
- When an organization's Wix plan subscription is cancelled, all members' Wix labels should be removed.
- When an organization downgrades to a different Wix plan tier, old tier labels should be replaced with new ones.

## Implementation notes

- The Wix Contacts API supports label removal — a `remove_labels` counterpart to the existing `add_labels` function in `squarelet/organizations/wix.py` would need to be implemented.
- Cleanup should be handled asynchronously via Celery tasks, similar to the existing `sync_wix` task.

allanlasser:
Closes #589

Adds missing "unsync" handling for when a subscription ends or a user leaves an org subscribed to a Sunlight plan.

### DIFF
--- a/squarelet/organizations/models/organization.py
+++ b/squarelet/organizations/models/organization.py
@@ -570,6 +570,14 @@ class Organization(AvatarMixin, models.Model):
             # modify the subscription
             self.subscription.modify(plan)
 
+        # Remove old Wix labels when plan is changing away from a wix plan
+        if (
+            from_plan
+            and from_plan.wix
+            and (not plan or not plan.wix or from_plan.pk != plan.pk)
+        ):
+            self._dispatch_wix_unsync(from_plan)
+
         self.change_logs.create(
             user=user,
             reason=ChangeLogReason.updated,
@@ -578,6 +586,38 @@ class Organization(AvatarMixin, models.Model):
             to_plan=plan,
             to_max_users=self.max_users,
         )
+
+    def _dispatch_wix_unsync(self, plan):
+        """Dispatch Wix unsync tasks for this org's users and, if this org is a
+        resource-sharing group, its member/child orgs.
+
+        Tasks are deferred until after the current transaction commits so that
+        workers see the post-change state when computing which labels a user
+        still qualifies for.
+        """
+        # pylint: disable=import-outside-toplevel
+        # Squarelet
+        from squarelet.organizations.tasks import (
+            unsync_wix,
+            unsync_wix_for_group_member,
+        )
+
+        org_pk, plan_pk = self.pk, plan.pk
+        user_pks = list(self.users.values_list("pk", flat=True))
+        member_pks = child_pks = []
+        if self.collective_enabled and self.share_resources:
+            member_pks = list(self.members.values_list("pk", flat=True))
+            child_pks = list(self.children.values_list("pk", flat=True))
+
+        def _dispatch():
+            for user_pk in user_pks:
+                unsync_wix.delay(org_pk, plan_pk, user_pk)
+            for member_pk in member_pks:
+                unsync_wix_for_group_member.delay(member_pk, org_pk, plan_pk)
+            for child_pk in child_pks:
+                unsync_wix_for_group_member.delay(child_pk, org_pk, plan_pk)
+
+        transaction.on_commit(_dispatch)
 
     def subscription_cancelled(self):
         """The subscription was cancelled due to payment failure"""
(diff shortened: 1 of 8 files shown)


squarelet/organizations/models/organization.py
```diff
@@ -570,6 +570,14 @@ def set_subscription(self, token, plan, max_users, user, payment_method=None):
             # modify the subscription
             self.subscription.modify(plan)
 
+        # Remove old Wix labels when plan is changing away from a wix plan
+        if (
+            from_plan
+            and from_plan.wix
+            and (not plan or not plan.wix or from_plan.pk != plan.pk)
```
mitchelljkotler: The condition in the parens reads as we unsync if:
1. there is no plan, ie they cancelled there membership
2. the new plan is not also a wix plan
3. they are switching plans

3 seems to subsume 2 - if they are switching from a wix plan to a non-wix plan, it will correctly unsync them.  If they are moving from a wix plan to another wix plan, condition 3 will still trigger and unsync them.  Is this what we want? (Will they get re-synced under the new plan?).  If it is, I think condition 2 can be removed.  If not, condition 3 needs to be more specific (I think combined with condition 2) or possibly removed (if the new plan is not a wix plan, it must be different then the from plan, since we check that it is a wix plan above)

mitchelljkotler (changes requested):

squarelet/organizations/models/organization.py
allanlasser: If switching between two Wix plans:

1. The user will be synced with details from the new plan (see 545–555 above).
2. We want to remove labels from the old plan from their record in Wix

For example, when upgrading from Essential to Enhanced we want to add the "Enhanced" label and remove the "Essential" label from each member of the org.

I think you're right that `not plan.wix` is irrelevant here. We want to remove the labels from the old plan regardless of whether the new plan syncs with Wix.

squarelet/organizations/wix.py
```diff
@@ -139,6 +171,61 @@ def sync_wix(organization, plan, user):
...
+        )
+        return
+
+    remove_labels(headers, contact_id, plan, label_keys=labels_to_remove)
+
+
+def get_wix_labels_for_user(user):
+    """Get all Wix labels a user qualifies for across all their memberships."""
+    labels = set()
+    for membership in user.memberships.select_related("organization___plan").all():
```
mitchelljkotler: Claude pointed out that this has a triple underscore instead of a double

mitchelljkotler (changes requested):
One more small fix then this looks good

squarelet/organizations/wix.py
```diff
@@ -139,6 +171,61 @@ def sync_wix(organization, plan, user):
...
+        )
+        return
+
+    remove_labels(headers, contact_id, plan, label_keys=labels_to_remove)
+
+
+def get_wix_labels_for_user(user):
+    """Get all Wix labels a user qualifies for across all their memberships."""
+    labels = set()
+    for membership in user.memberships.select_related("organization__plan").all():
```
mitchelljkotler: Sorry, I just took a closer look at this and realized what was going on here.  There is a deprecated `_plan` field, which is what you were loading, which is why there was a triple underscore.  That field is not used though, and does not need to be loaded.  There is no `plan` field - there is a `plans` many to many relationship.  `.plan` is a property method which just returns the first plan from `plans` since we currently only support one plan.  On the Stripe branch, `_plan` will be removed, and we will start supporting multiple plans.  I think this means we want to `prefetch_related("organizations__plans")`, which will pre-load the M2M relation, then walk through all plans `for plan in org.plans.all()` - for two reasons: 1. If you you `org.plan` it will not use the pre-loaded plans since it is a different query (it uses `.first()`) 2. We will support multiple plans soon.

mitchelljkotler (changes requested):

allanlasser:
@mitchelljkotler I fixed al the failing tests and addressed your earlier feedback on plan access. Ready for a last look here.

mitchelljkotler (approved):

PR #647, opened 2026-04-13
Update the Stripe SDK to the latest version and the Stripe API to the latest version

mitchelljkotler:
Closes #644 

This needs heavy testing on the review app to ensure new API calls function correctly


duckduckgrayduck:
Not within scope of this PR bc I think the payments frontend is on our goals to redesign shortly, one thing to note is that it is clunky that I have to remove a card on file first and then add a new card from within billing settings for it to stick. Approved PR with understanding checks will be fixed

duckduckgrayduck (approved):

PR #651, opened 2026-04-14
Only show org admin list to authenticated users

allanlasser:
This makes it harder for bad actors to identify organization admins for targeting in social engineering attacks.

### DIFF
--- a/e2e/organizations.spec.ts
+++ b/e2e/organizations.spec.ts
@@ -64,10 +64,10 @@ test.describe("Organization Viewing", () => {
       expect(response?.status()).toBe(404);
     });
 
-    test("sees only admins in member list", async ({ page }) => {
+    test("cannot see the org's member list", async ({ page }) => {
       await page.goto("/organizations/e2e-public-org/");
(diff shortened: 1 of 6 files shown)


duckduckgrayduck:
This looks good, just needs the e2e test fixed to match new behavior. 

 I noticed one other thing on production that we should also address. 

- Someone's 2FA enrollment tag should only visible to admins of an org, no one else

duckduckgrayduck:
To be a little more guarded against spam collection, since account creation is a low barrier- do we want to hide admin emails for users that haven't confirmed their email address? 

We could provide a  like "Confirm your email address to see email addresses of admins"

I ask because in the user search branch which we will merge shortly, our barrier for showing email addresses in user search is much higher (be a part of a verified org). 

Everything else appears to function as it should

allanlasser:
I think that's a good suggestion, but I'm inclined to support it in a follow-on PR so we can get this change in. I'll make a new issue for it.

duckduckgrayduck:
Definitely not a blocker, this looks great already!

allanlasser:
https://github.com/MuckRock/squarelet/issues/652

PR #656, opened 2026-04-20
Restrict invitation actions to users with confirmed emails

### LINKED ISSUE #619: Require a verified email address to become an org admin

allanlasser:
Resolves #619

Users must have confirmed their email addresses before they can accept an invitation to join an org **as an admin**.

This also relaxes the restriction allowing uses to see invitations and send requests if they have an unverified email address for a consistent user experience. It's acceptable for users without confirmed emails to be able to accept or request org membership, given our other guardrails to prevent abuse. This results in a smoother first-run experience for users.

Improves code quality by rendering all invitation mutation UI from a  `InvitationAcceptForm` that also validates the POST data. Forms are bound to invitations during context generation, so we render a unique form for each invitation in the template.

### DIFF
--- a/e2e/organizations.spec.ts
+++ b/e2e/organizations.spec.ts
@@ -465,8 +465,9 @@ test.describe("Member Management", () => {
     const anonContext = await browser.newContext({ ignoreHTTPSErrors: true });
     const anonPage = await anonContext.newPage();
     await anonPage.goto(`https://dev.squarelet.com${invitationPath}`);
-    await expect(anonPage.locator("a:has-text('Sign Up')")).toBeVisible();
-    await expect(anonPage.locator("a:has-text('Log In')")).toBeVisible();
+    const controls = anonPage.locator(".control-group");
+    await expect(controls.locator("a:has-text('Sign Up')")).toBeVisible();
+    await expect(controls.locator("a:has-text('Log In')")).toBeVisible();
     await anonContext.close();
 
     // Verify the link invitation shows "Copy link" (not "Resend") on manage-members
--- a/e2e/organizations.spec.ts
+++ b/e2e/organizations.spec.ts
@@ -1074,4 +1074,71 @@ test.describe("Invitations for unverified users", () => {
     await expect(invitation).toBeVisible();
     await expect(invitation.locator('button[value="accept"]')).toBeVisible();
   });
+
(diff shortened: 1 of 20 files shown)


duckduckgrayduck:
Someone who currently is an admin of an org and has an unverified email, should they be able to send invitations? Unclear to me from the description in the PR. Currently, I have tested, an admin without any verified emails is able to create invitations.

allanlasser:
It's a really good question what to do for existing users. I think we keep the behavior as-is for sending invitations here, since this set of changes is focused on limiting the ability to accept invitations.

duckduckgrayduck:
When I click "Verify your email address" In the invitation- I get an alert "Invalid choice" and it doesn't let me move forward

allanlasser:
Ah, good catch. It must be posting to the wrong endpoint.

allanlasser:
@duckduckgrayduck It was a templating issue with the form. I cleaned up the `invitation_detail.html` template and styling a bit:

Screenshot 2026-04-27 at 21 19 13

duckduckgrayduck:
User who doesn't have a verified email address but has an account receives an invite to become an admin of an org -> Shows me a button to "Verify your email address" which doesn't do anything at the moment. Doesn't fail, doesn't succeed, just does nothing

allanlasser:
@duckduckgrayduck In response to your feedback, I replaced a form with a link to the user detail page's "connected accounts" section, where they can trigger an email confirmation flow. I also fixed failing E2E tests and fixed conflicts against `master` (this branch was about 120 commits behind!).

duckduckgrayduck:
I send an invite to a user with no verified emails to be a regular member. When logged in as that verified user, I do not see the invitations when logged in (I should, as long as it is an invite to be a member, and not an admin). I can go into the email and accept the invitation to be a member though. 

Inviting an unverified user as an admin all is correct in the flow. Allauth rate limits the confirmation email sending I think, which is why it sometimes silently does nothing and other times does actually send the email just fyi

allanlasser:
@duckduckgrayduck now invitations appear to all users regardless of email confirmation status, while this branch's consolidation onto the `InvitationAcceptForm` and its template ensures consistent handling across the app. The template conditionally renders the "Confirm email" link when needed, and the form validates whether the invitation can be accepted.

duckduckgrayduck (approved):
Verified all invitation streams for unverified users works as intended.

PR #659, opened 2026-04-24
Add PlanFilter to Organization admin

allanlasser:
Helps us answer "How many orgs are subscribed to plan X?" with an update to the Django Admin.

Lets staff slice the Organization list view by the Plan an org is subscribed to, plus a "No plan" option for orgs with no Subscription rows. Cancelled subscriptions still count as subscribed, matching the semantics of Organization.plan.

Co-authored with Claude Opus 4.7

### DIFF
--- a/squarelet/organizations/admin.py
+++ b/squarelet/organizations/admin.py
@@ -1,6 +1,6 @@
 # Django
 from django.contrib import admin, messages
-from django.db.models import Count, JSONField, Q, Sum
+from django.db.models import Count, Exists, JSONField, OuterRef, Q, Sum
 from django.forms.models import BaseInlineFormSet
 from django.forms.widgets import Textarea
 from django.http.response import HttpResponse
@@ -325,6 +325,14 @@ class OrganizationAdmin(VersionAdmin):
         "get_subtypes",
         "get_outstanding_invoices",
     )
+
+    def get_list_display(self, request):
+        """Add cancelled column when filtering by plan"""
+        list_display = list(super().get_list_display(request))
+        if "plan" in request.GET and request.GET["plan"] != "none":
+            list_display.append("get_subscription_renews")
+        return list_display
+
     list_filter = (
         PlanFilter,
         "individual",
@@ -394,7 +402,7 @@ class OrganizationAdmin(VersionAdmin):
     )
 
     def get_queryset(self, request):
-        return (
+        qs = (
             super()
             .get_queryset(request)
             .prefetch_related("subtypes", "domains")
@@ -407,6 +415,18 @@ class OrganizationAdmin(VersionAdmin):
                 ),
             )
         )
+        plan_value = request.GET.get("plan")
+        if plan_value and plan_value != "none":
+            qs = qs.annotate(
+                subscription_cancelled=Exists(
+                    Subscription.objects.filter(
+                        organization_id=OuterRef("pk"),
+                        plan_id=plan_value,
+                        cancelled=True,
+                    )
+                )
+            )
+        return qs
 
     def save_model(self, request, obj, form, change):
         super().save_model(request, obj, form, change)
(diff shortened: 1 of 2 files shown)


duckduckgrayduck:
Seems to work well- I do feel though that this would be greatly improved if we could mark who is currently subscribed vs. cancelled

allanlasser:
I weighed that—orgs in a cancelled state will continue to have access to a plan until their subscription is deleted at the end of the term. It feels more accurate to include cancelled subscriptions in these groups, so they represent active subscriptions.

I think adding a column to this view would be the best way to identify orgs that won't have their subscription renew, and that's a good thing to add here.

allanlasser:
Now when filtering by a plan, a column's added to indicate whether the subscription will renew.

duckduckgrayduck:
The number of SQL queries (as viewed in the Django debug toolbar) with this new column in the view radically jumped from ~30ish to ~200. This needs optimization

allanlasser:
@duckduckgrayduck Take another look here? SQL optimization isn't my strong suit, so any advice here is welcome.

squarelet/organizations/admin.py
```diff
@@ -186,6 +186,26 @@ class MembershipsInline(admin.TabularInline):
...
+        plans = Plan.objects.order_by("name").values_list("pk", "name")
+        return [("none", "— No plan —"), *plans]
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        if value is None:
+            return queryset
+        if value == "none":
+            return queryset.filter(subscriptions__isnull=True)
+        return queryset.filter(subscriptions__plan_id=value)
```
mitchelljkotler, line 206: This does the same thing as the subquery but is both more performant and much clearer to read

squarelet/organizations/admin.py
```diff
@@ -386,6 +415,16 @@ def get_queryset(self, request):
                 ),
             )
         )
+        plan_value = request.GET.get("plan")
+        if plan_value and plan_value != "none":
+            qs = qs.prefetch_related(
```
mitchelljkotler, line 420: Doing a prefetch is again more perofrmant and clearer

squarelet/organizations/admin.py
```diff
@@ -437,6 +476,15 @@ def get_outstanding_invoices(self, obj):
     get_outstanding_invoices.short_description = "Outstanding Invoices"
     get_outstanding_invoices.admin_order_field = "outstanding_invoice_count"
 
+    def get_subscription_renews(self, obj):
+        subs = getattr(obj, "plan_subscriptions", None)
+        if subs is None:
+            return None
+        return not any(s.cancelled for s in subs)
```
mitchelljkotler, line 483: We need to check for cancelled here since it was moved out of the query

squarelet/organizations/admin.py
```diff
@@ -437,6 +476,15 @@ def get_outstanding_invoices(self, obj):
     get_outstanding_invoices.short_description = "Outstanding Invoices"
     get_outstanding_invoices.admin_order_field = "outstanding_invoice_count"
 
+    def get_subscription_renews(self, obj):
+        subs = getattr(obj, "plan_subscriptions", None)
+        if subs is None:
+            return None
```
mitchelljkotler, line 482: Since this is marked as a boolean, returning a string will render it as a green check.  None will render empty.

squarelet/organizations/admin.py
```diff
@@ -437,6 +476,15 @@ def get_outstanding_invoices(self, obj):
...
+        return not any(s.cancelled for s in subs)
+
+    get_subscription_renews.short_description = "Will Renew"
+    get_subscription_renews.boolean = True
```
mitchelljkotler, line 486: We lose the ability to order by cancellations, but this seems like an acceptable trade off for the performance implications

allanlasser:
@mitchelljkotler thanks for the optimizations here! They all make sense. Think this is ready to go on Monday?

mitchelljkotler (approved):

PR #665, opened 2026-04-29
Bump web worker dynos to 2x to attempt to make it less lean on deploy…

duckduckgrayduck:
… previews

### DIFF
--- a/app.json
+++ b/app.json
@@ -5,8 +5,8 @@
   "repository": "https://github.com/muckrock/squarelet",
   "formation": {
     "web": {
-      "quantity": 2,
-      "size": "standard-1x"
+      "quantity": 1,
+      "size": "standard-2x"
     },
     "beat": {
       "quantity": 1,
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -5,7 +5,7 @@
   "repository": "https://github.com/muckrock/squarelet",
   "formation": {
     "web": {
-      "quantity": 1,
+      "quantity": 2,
       "size": "standard-1x"
```
allanlasser: I think we want to use 1 `standard-2x` dyno here (1GB memory), not 2 `standard-1x` dynos (512MB).

allanlasser (changes requested):

app.json
duckduckgrayduck: I've just been bumping the number of dynos, not their types when I've had issues, but you're probably right

allanlasser (approved):

PR #680, opened 2026-05-19
Support arbitrary, temporary Entitlement grants

### LINKED ISSUE #666: Support arbitrary, temporary Entitlement grants
## Problem

For some of our ecosystem partnerships, we want to extend entitlements to 3rd party apps for limited periods, to a limited number of newsrooms.

For example:

- 15 organizations of our choice get pilot access to a service for a few months
- all verified newsrooms with a paid subscription get promotional access to a service for a few months

## Solution

After discussing this with Mitch, I think the right design for this is an `EntitlementGrant` model with properties for:

- A set of entitlements to grant (M2M with Entitlements)
- An explicit set of organizations to receive the grant (M2M with organizations)
- An implicit set of organizations to receive the grant, based on a set of predefined parameters:
    - verified status
    - active subscription

Each rule could be something like a lambda function that accepts an organization, though another approach may be better.

When serializing user data through the API, we'll want to collate/union any available entitlements per organization, so that we combine and deduplicate:
- entitlements received through a subscription
- entitlements received through a grant

allanlasser:
Fixes #666

Adds a new `EntitlementGrant` model that lets us give Entitlements to Organizations. 

Previously, the only way to associate Entitlements with Organizations was through  a Subscription to a Plan.

This allows us to grant access to resources in an arbitrary way to any org, and revoke them as needed, too.

This uses our same code path for refreshing entitlements on a monthly cadence.

This also updates our serializer to union entitlements, which will come in handy when we support multiple concurrent subscriptions.


squarelet/organizations/models/payment.py
```diff
@@ -762,6 +766,136 @@ def public(self):
...
+        explicit_q = Q(entitlement_grants=self)
+
+        rule_clauses = []
+        if self.require_verified:
+            rule_clauses.append(Q(verified_journalist=True))
+        if self.require_active_subscription:
+            # Mirrors org.has_active_subscription() = bool(subscriptions.first())
+            rule_clauses.append(Q(subscriptions__isnull=False))
+
+        if rule_clauses:
```
mitchelljkotler, line 891: Checking my understanding:

If you check `require_verified` then this entitlement grant applies to all Organizations which are verified (filtered by individual/group as appropriate)

If you check `require_active_subscription` then it applies to all Orgs with an active subscription (again filtered by individual/group)

If both are checked, then it applies to all Orgs which are both verified and have an active subscription.

squarelet/organizations/models/payment.py
```diff
@@ -762,6 +766,136 @@ def public(self):
...
+            rule_q = rule_clauses[0]
+            for clause in rule_clauses[1:]:
+                rule_q &= clause
+            return eligible.filter(explicit_q | rule_q).distinct()
+        return eligible.filter(explicit_q).distinct()
```
mitchelljkotler: if there is only `explicit_q` then `distinct` is not necessary as you can't add the same org more than once.

squarelet/organizations/models/payment.py
```diff
@@ -762,6 +766,136 @@ def public(self):
...
+    objects = EntitlementGrantQuerySet.as_manager()
+
+    class Meta:
+        ordering = ("-created_at", "name")
+
+    def __str__(self):
+        return self.name
+
+    def save(self, *args, **kwargs):
+        if self.update_on is None:
```
mitchelljkotler: If you are just trying to set a default value, you can do this more cleanly by setting the default to a callable:

https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.Field.default

squarelet/organizations/models/payment.py
```diff
@@ -762,6 +766,136 @@ def public(self):
...
+    def matches(self, org):
+        if not self.active:
+            return False
+        # Org-type filter applies to both explicit and rule-based matches.
+        if org.individual and not self.for_individuals:
+            return False
+        if not org.individual and not self.for_groups:
+            return False
+        # Uses `.all()` so a prefetched `organizations` relation is reused.
+        if any(o.pk == org.pk for o in self.organizations.all()):
```
mitchelljkotler: `if self.organizations.filter(pk=org.pk).exists():`

squarelet/organizations/querysets.py
```diff
@@ -164,6 +164,50 @@ def get_owned(self, user):
...
+        """Return the deduped union of plan-derived and grant-derived entitlements
+        currently available to `org`. Optionally scoped to a single OIDC client."""
+        # Lazy import to avoid a circular import (payment.py imports this module)
+        # pylint: disable=import-outside-toplevel
+        # Squarelet
+        from squarelet.organizations.models.payment import EntitlementGrant
+
+        plan_q = Q(plans__organizations=org)
+        grant_pks = [g.pk for g in EntitlementGrant.objects.matching(org)]
+        grant_q = Q(grants__in=grant_pks)
```
mitchelljkotler: This can be `grant_q = Q(grants__in=EntitlementGrant.objects.matching(org))`

Django will automatically use the PK if you pass in a model object

mitchelljkotler (changes requested):
I created another PR with SQL optimizations

squarelet/organizations/models/payment.py
allanlasser: Correct

allanlasser:
@mitchelljkotler I've addressed your feedback and incorporated your fixes!

mitchelljkotler (approved):

PR #685, opened 2026-05-21
optimize fuzzy searching memory usage

mitchelljkotler:
Does not load full organization objects into memory for fuzzy, only grab the ID and name


duckduckgrayduck:
The e2e test is failing and fuzzy search on org creation is broken. because of this:
ValueError
Level: Error
chunk_size must be provided when using QuerySet.iterator() after prefetch_related().
This is a Django limitation. We may have to just remove iterator()

Issue #686, opened 2026-05-27
Organization.MultipleObjectsReturned Error


duckduckgrayduck:
Sentry error:
https://muckrock.sentry.io/issues/7257289224/events/e0d5223860d945b2ab9bf49545d13180/?project=1422309&query=is%3Aunresolved&referrer=previous-event

Offending commit:
https://github.com/MuckRock/squarelet/commit/e85823a042be9b9062008b75d5c19a8e7f940fcb

PR #687, opened 2026-05-27
Add a distinct on consent filter

duckduckgrayduck:
Testing a fix for #686 
Co-authored with Claude

### DIFF
--- a/squarelet/organizations/viewsets.py
+++ b/squarelet/organizations/viewsets.py
@@ -41,7 +41,7 @@ class OrganizationViewSet(viewsets.ModelViewSet):
                 return self.queryset.filter(
                     users__userconsent__client=client,
                     users__userconsent__expires_at__gt=timezone.now(),
-                )
+                ).distinct()
         return self.queryset
 
 


duckduckgrayduck:
Tested in the Django shell of the deploy preview, it fixes the issue. 
>>

from django.utils import timezone
from datetime import timedelta
from django.db.models import Count
from oidc_provider.models import Client, UserConsent
from squarelet.organizations.models import Organization

# Get a client with require_consent=True
client = Client.objects.filter(require_consent=True).first()  # "Zapier"

# Find an org with multiple users
org = Organization.objects.annotate(n=Count('users')).filter(n__gt=1).first()

# Seed consent records for each user
for user in org.users.all():
    UserConsent.objects.get_or_create(
        user=user,
        client=client,
        defaults={
            "expires_at": timezone.now() + timedelta(days=30),
            "scope": [],
            "date_given": timezone.now(),
        }
    )

# Reproduce the bug
Organization.objects.filter(
    users__userconsent__client=client,
    users__userconsent__expires_at__gt=timezone.now(),
).get(uuid=org.uuid)  # raises MultipleObjectsReturned **--> This is the bug.** 

# Verify the fix
Organization.objects.filter(
    users__userconsent__client=client,
    users__userconsent__expires_at__gt=timezone.now(),
).distinct().get(uuid=org.uuid)  # returns <Organization: Luminant Media>

allanlasser (approved):

PR #690, opened 2026-05-27
Update payment handling data model

### LINKED ISSUE #648: Redo Stripe data model
Use webhooks to cache all data locally

mitchelljkotler:
This caches stripe data locally, to avoid needing to go to the Stripe API whenever that data is needed.
What's being cached:
Customer     │ payment_brand, payment_last4, payment_exp_month, payment_exp_year, stripe_payment_method_id
Subscription │ stripe_status, current_period_end
Invoice      │ hosted_invoice_url

The data is updated when Stripe sends webhooks for these models.

Testing - can test general payment related functionality and ensure everything displays as expected.


squarelet/organizations/models/payment.py
```diff
@@ -51,10 +49,18 @@ class Customer(models.Model):
...
     )
 
+    card_brand = models.CharField(max_length=20, blank=True, default="")
+    card_last4 = models.CharField(max_length=4, blank=True, default="")
+    card_exp_month = models.PositiveSmallIntegerField(null=True, blank=True)
+    card_exp_year = models.PositiveSmallIntegerField(null=True, blank=True)
+    stripe_payment_method_id = models.CharField(
+        max_length=255, blank=True, default=""
+    )
+
```
allanlasser: Let's break this out into its own table, so that we can support multiple cards for customers. This seems worth the performance tradeoff for the additional JOIN.

requirements/base.txt
allanlasser, line 313: Current Stripe library is v15.2.1, is there a reason to lock to 15.0.0?

squarelet/organizations/models/payment.py
mitchelljkotler: I agree with breaking this out into its own table, but will do that in a follow up PR

requirements/base.txt
mitchelljkotler: I think 15.0.0 was the latest when I did the update - I am not going to update for every point release.  This is now in our tracker for things to update, so we will update it regularly.

eyeseast:
Got a 404 on plan detail, here locally: https://dev.squarelet.com/plans/2-professional/. It links to the organization, even if I'm paying as an individual. (This may already be happening in production, but if so, we should fix it here.)

eyeseast (approved):

allanlasser (approved):

PR #691, opened 2026-05-27
Update payment handling multiple subscriptions

mitchelljkotler:
Back end support for organizations to have multiple subscriptions.

Removes assumptions that an organization can only have one subscription.  Anchors all subscriptions to the same day of the month.

This functionality is not exposed yet through the UI, so actually creating multiple subscriptions is not possible yet.


allanlasser:
I noticed this is still based on `update-payment-handling-payment-intents` instead of `master`. That may help with some of the file conflicts reported here.

squarelet/organizations/models/organization.py
```diff
@@ -517,80 +470,120 @@ def member_users(self, request=None):
...
+        for this plan.
+        """
+        if self.subscriptions.filter(plan=plan).exists():
+            raise ValueError(
+                f"Organization already has an active subscription to {plan}"
+            )
+
+        is_first = not self.subscriptions.exists()
+
+        self.max_users = max_users
```
allanlasser: Adding a new subscription with a different `max_users` value will override the previous org-wide value. Will this impact other subscriptions?

I'm going to look at our callsites for add_subscription to sef if we can eliminate the max_users arg.

squarelet/organizations/models/organization.py
```diff
@@ -517,80 +470,120 @@ def member_users(self, request=None):
...
+            customer.email = self.email
+            customer.save()
+
+        # update_on must be None when start() is called so the first subscription
+        # receives no billing_cycle_anchor (Stripe sets its own anchor). Only after
+        # the subscription exists do we record the anchor for subsequent subscriptions
+        # to align to.
+        self.subscriptions.start(
+            organization=self, plan=plan, payment_method=payment_method
+        )
```
allanlasser, line 523: Making sure I'm following the logic here. The first subscripion added to an organization uses the anchor date set by Stripe—checks out. Two questions lead from this:
1. Why are we setting our own `update_on` for the billing anchor date, instead of using the one provided by Stripe?
2. Shouldn't any subsequent subscriptions get the `update_on` anchor date passed through to them?

squarelet/organizations/models/organization.py
```diff
@@ -517,80 +470,120 @@ def member_users(self, request=None):
...
+        # pylint: disable=import-outside-toplevel
+        # Squarelet
+        from squarelet.organizations.models.payment import Subscription as Sub
+
+        if isinstance(plan_or_subscription, Sub):
+            sub = plan_or_subscription
+        else:
+            sub = self.subscriptions.get(plan=plan_or_subscription)
+
+        cancelled_plan = sub.plan if sub.plan and sub.plan.wix else None
```
allanlasser: Since this is primarily about wix sync dispatch, a clearer name would help here.

squarelet/organizations/models/organization.py
```diff
@@ -517,80 +470,120 @@ def member_users(self, request=None):
...
+        """Cancel the subscription for the given plan or Subscription instance."""
+        # pylint: disable=import-outside-toplevel
+        # Squarelet
+        from squarelet.organizations.models.payment import Subscription as Sub
+
+        if isinstance(plan_or_subscription, Sub):
+            sub = plan_or_subscription
+        else:
+            sub = self.subscriptions.get(plan=plan_or_subscription)
+
```
allanlasser, line 572: This doesn't handle an edge case where a subscription can't be found and `sub` ends up as `None`. This would throw errors when trying to create the changelog entry and call `sub.cancel()`.

squarelet/organizations/models/organization.py
```diff
@@ -629,20 +622,24 @@ def subscription_cancelled(self, subscription=None):
 
         Args:
             subscription: The specific subscription to cancel. If None,
-                cancels self.subscription (the org's current subscription).
+                cancels self.subscriptions.first() (the org's current subscription).
         """
-        subscription = subscription or self.subscription
+        subscription = subscription or self.subscriptions.first()
```
allanlasser: If no subscription is found, we should log something and escape. That would avoid a crash when calling `subcription.cancel()` below, and avoid the need for conditional checks elsewhere in this method.

squarelet/organizations/models/organization.py
```diff
@@ -629,20 +622,24 @@ def subscription_cancelled(self, subscription=None):
...
```
allanlasser: It feel _really risky_ to me to default to the first subscription found here. If we can't find a subscription to cancel before calling into this function, we should probably raise an error at the call-site.

It looks like the place this gets called without a subscription is `handle_invoice_failed`. If we have an invoice with no affiliate subscription, we should probably prompt some kind of manual intervention.

squarelet/organizations/models/membership.py
```diff
@@ -105,15 +114,17 @@ def delete(self, *args, **kwargs):
...
         group_wix_plans = [(g.pk, p.pk) for g, p in org.get_wix_plans_from_groups()]
 
         with transaction.atomic():
             super().delete(*args, **kwargs)
             transaction.on_commit(lambda: send_cache_invalidations("user", user_uuid))
-            if direct_plan_pk is not None:
+            for plan_pk in direct_plan_pks:
                 transaction.on_commit(
-                    lambda: unsync_wix.delay(org.pk, direct_plan_pk, user_pk)
+                    lambda p=plan_pk: unsync_wix.delay(org.pk, p, user_pk)
```
allanlasser, line 127: Why do we not pass through org.pk and user_pk as arguments here, when we do that in the lambda function in `save`?

squarelet/organizations/models/payment.py
```diff
@@ -235,15 +242,25 @@ def start(self, payment_method="card"):
...
-                "verified_premium_org"
-                if self.organization.verified_journalist
-                else "unverified_premium_org"
+            already_has_org_entitlement = (
+                self.organization.subscriptions.exclude(pk=self.pk)
+                .filter(
+                    plan__entitlements__slug="organization",
+                )
+                .exists()
             )
```
allanlasser, line 266: Why are we locking our Mailchimp journey trigger to a hardcoded entitlement slug?

squarelet/payments/forms.py
```diff
@@ -162,7 +162,6 @@ def _configure_organization_field(self):
             if self.plan:
                 subscribed_orgs = Organization.objects.filter(
                     subscriptions__plan=self.plan,
-                    subscriptions__cancelled=False,
                 )
```
allanlasser, line 166: Is this change to support re-subscribing to a cancelled, but still active, plan?

squarelet/organizations/forms.py
```diff
@@ -129,7 +129,7 @@ def _set_group_options(self):
         # on a private plan, plus private plans they have been given access to
         plans = Plan.objects.choices(self.organization)
         self.fields["plan"].queryset = plans
-        self.fields["plan"].default = self.organization.plan
+        self.fields["plan"].default = self.organization.plans.first()
```
allanlasser, line 132: Pretty big flag that this form continues to assume orgs will only have one active subscription. After this change, it'll be possible for orgs to have multiple active subscriptions, if they purchase through our `PlanPurchaseForm` (`squarelet/payments/forms.py`).

squarelet/organizations/views/create.py
```diff
@@ -23,7 +23,7 @@ def form_valid(self, form):
         organization.change_logs.create(
             reason=ChangeLogReason.created,
             user=self.request.user,
-            to_plan=organization.plan,
+            to_plan=organization.plans.first(),
```
allanlasser, line 26: The assocaited form has no support for subscribing an org to a plan. Why are we including `organization.plans.first()` here?

squarelet/organizations/views/subscription.py
```diff
@@ -57,18 +57,29 @@ class UpdateSubscription(OrganizationPermissionMixin, UpdateView):
...
+        max_users = form.cleaned_data.get("max_users")
+
+        if not current_plan and new_plan:
+            organization.add_subscription(new_plan, max_users, user, token=token)
+        elif current_plan and not new_plan:
+            organization.remove_subscription(current_plan, user)
+        elif current_plan and new_plan:
+            if token:
+                organization.save_card(token, user)
+            organization.modify_subscription(current_plan, new_plan, max_users, user)
```
allanlasser, line 74: Do we want to continue to maintain the `modify` codepath, when upgrading/switching plans could be handled by a `remove_` + `add_`? Is there a strong case for a modification of the current subscription in-place?

squarelet/organizations/views/subscription.py
```diff
@@ -131,7 +142,7 @@ def get_context_data(self, **kwargs):
 
     def get_initial(self):
         return {
-            "plan": self.object.plan,
+            "plan": self.object.plans.first(),
```
allanlasser: Flagging that this assumes we have just 1 subscription on our account that we're upgrading.

squarelet/payments/views.py
```diff
@@ -116,7 +116,7 @@ def get_context_data(self, **kwargs):
             # Check user's individual organization
             individual_org = user.individual_organization
             individual_subscription = individual_org.subscriptions.filter(
-                plan=plan, cancelled=False
+                plan=plan
             ).first()
```
allanlasser, line 121: Shouldn't we be including all subscriptions for an individual or group in our existing subscriptions list? This isused in the template to check whether the current plan can be purchased.

squarelet/organizations/migrations/0060_remove_organization_subscription_id.py
```diff
@@ -0,0 +1,15 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("organizations", "0059_alter_organization_options"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="organization",
+            name="subscription_id",
```
allanlasser, line 13: Do we need to look for an existing subscription matching this ID? Does this contain any relevant data?

squarelet/organizations/migrations/0064_alter_entitlementgrant_update_on.py
```diff
@@ -1,7 +1,12 @@
 # Generated by Django 5.2.12 on 2026-05-26 18:43
 
-import squarelet.organizations.models.payment
 from django.db import migrations, models
+from django.utils import timezone
+from dateutil.relativedelta import relativedelta
+
+
+def _default_grant_update_on():
+    return timezone.now().date() + relativedelta(months=1)
```
allanlasser, line 9: This is defined in two different places in the changeset. If this ever changes, it'll diverge between our migration and our code. Do we want to lock this in here?

squarelet/organizations/migrations/0067_remove_entitlementgrant_update_on.py
```diff
@@ -0,0 +1,16 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("organizations", "0066_org_update_on_from_subscription"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="entitlementgrant",
+            name="update_on",
```
allanlasser, line 14: This is never copied onto the org and is simply lost. This could create orgs with an entitlement grant but no subscriptions and no `update_on` value.

squarelet/organizations/tasks.py
```diff
@@ -40,25 +41,48 @@ def restore_organization():
...
-    subscriptions.filter(cancelled=True).delete()
-    subscriptions.update(update_on=today + Interval("1 month"))
-
-    # --- Entitlement grants ---
-    # Snapshot the matching orgs for each active expired grant before bumping
-    # update_on, then bulk-bump in one query. Inactive expired grants get the
-    # bump too but contribute no UUIDs to the broadcast.
-    expired_grants = EntitlementGrant.objects.expired(today)
+    due_org_ids = list(
+        Organization.objects.filter(update_on__lte=today).values_list("id", flat=True)
```
allanlasser, line 45: I am still very skeptical of this design—conflating the **subscription billing anchor date** with the **entitlement refresh date** still feels like we're combining two seperate concerns into a single value. It's a kind of false simplicity which  _actually more complicated to reason about_.

The field here is doing double duty, while the nullable fallback to a synthetic hardcoded value.

This also assumes we'll only ever have entitlements that refresh monthly—as we become more a platform provider for 3rd party ecosystems projects, this contrains our flexibility to adapt to apps that may want a monthly, daily, or even hourly refresh cycle (for example, a program with 5h usage resets like Claude Code). There's a tough balance to strike here—how do we make sure what we're building is flexible enough to meet needs we can't anticipate right now? We've used our own needs as the sole determiner, but that is incompatible with our goals for Accounts as the hub of an ecosystem of apps.

I'm still going to advocate for storing a relationship between an Organization and each of its entitlements, with a refresh date set on that relationship and a reference to where the relationship started from.

squarelet/organizations/migrations/0067_remove_entitlementgrant_update_on.py
allanlasser: Now that I understand we're simply hardcoding the `update_on` date for EntitlementGrants, this makes more sense. But that could be made more explicit through comments, README updates, documentation, etc..

squarelet/organizations/migrations/0064_alter_entitlementgrant_update_on.py
```diff
@@ -16,7 +21,7 @@ class Migration(migrations.Migration):
             name="update_on",
             field=models.DateField(
                 blank=True,
-                default=squarelet.organizations.models.payment.default_grant_update_on,
+                default=_default_grant_update_on,
```
allanlasser, line 24: Just realized that this updates the value fo the field, only to delete it later. Feels unnecessary!

squarelet/organizations/migrations/0060_remove_organization_subscription_id.py
mitchelljkotler: This field was deprecated, the subscription ID had previously been moved to the subscription model

squarelet/organizations/migrations/0064_alter_entitlementgrant_update_on.py
mitchelljkotler: This is an old migration - this function was removed from the code, which would break the migration.  Therefore, the function was moved inline to the migration so that it will work on new databases.  The field is later removed, but this is still needed to keep migrations working, unless we squash them.

squarelet/organizations/migrations/0064_alter_entitlementgrant_update_on.py
mitchelljkotler: See previous comment

squarelet/organizations/models/membership.py
mitchelljkotler: The one in `save` could be updated to be like this one.

This is an odd behavior in Python, which is better explained here:
https://docs.python-guide.org/writing/gotchas/#late-binding-closures

The short version is, only the variables that change in the loop (here plan_pk) need to be passed through the lambda.  When writing `save` I was overzealous in adding them all - this doesn't change anything besides being noisy.

squarelet/organizations/models/organization.py
mitchelljkotler: This is a good catch.  `max_users` is a property of the subscription now, not the organization.  Ideally, we will soon be retiring this concept, but it needs to keep working for now.  I have pushed a new commit which moves this value to `subscription.quantity`.

squarelet/organizations/models/organization.py
mitchelljkotler: 1. Stripe should always set a date 30 days out, so these should be equivalent, but using the time stamp it returns is more correct, so updating it to do that

2. The initial subscription will have `organization.update_on` set to `None`, so stripe will pick the anchor date.  Subsequent calls get `update_on` from the `organization` property, so it is passed in.

squarelet/organizations/models/organization.py
mitchelljkotler: The only call path for this currently loads the plan from the DB immediately before calling this, so not finding it should not be an issue.  Also, sub will not end up as `None`, `get` will throw a `DoesNotExist` error, which could be caught if the view changes in a way that makes this possible (which it will probably need to once the frontend supports multiple subscriptions properly).

squarelet/organizations/models/organization.py
mitchelljkotler: Agreed

squarelet/organizations/models/organization.py
allanlasser: OK, I now see that in `squarelet/organizations/querysets.py::SubscriptionQuerySet.start`, we're using `organization.update_on` as the `billing_anchor_date` value, which explains why we don't pass it through and why it's getting set only after calling `.start` on the first subscription added to the org.

Looking at this code as a whole, these org methods are indirect and non-obvious, making it harder to trace how the logic of a subscription starts. I think my new question here is why do we have these organization methods for adding and removing subscriptions on the Organization, if the Subscription queryset is managing the relationship between a plan and an organization? To me, it'd be clearer and remove a level of indirection to create the subscription from the Subscription queryset—all the data we access in here is passed to that start method.

So, instead of calling `organization.add_subscription(plan, max_users, user, token=None, payment_method=None)` and then it calling into `self.subscriptions.start(organization=self, plan=plan, payment_method=payment_method)`, we could call `Subscriptions.start(organization, plan, user, token=None, payment_method=None)` without any sub-calls. We'd start a new Subscription, which joins an Org to a Plan. Clear and easy to trace.

squarelet/organizations/models/organization.py
mitchelljkotler: Agreed with both here - this was falling through on previous behavior.  Updated.

squarelet/organizations/models/payment.py
mitchelljkotler: This is how it was originally spec'ed out - I agree it isn't inline with how we want to go moving forward.  As this PR is already larger than I would like, I think this fix can go into a follow up PR, as this didn't actually change in this PR.

My proposed solution to this would be entitlements (or plans?) gain a field for specifying mailchimp journeys

squarelet/organizations/views/create.py
mitchelljkotler: Seems like a fine safe guard in case that form changes to allow for plans.

squarelet/organizations/views/subscription.py
mitchelljkotler: If upgrading or downgrading a plan, this updates it in place on Stripe, which preserves the billing anchor and prorates properly based on how much time was left on the pay period.  A remove and add would mark the current subscription to be cancelled at the end of the pay period, and charge the user immediately for the new plan.

squarelet/organizations/views/subscription.py
mitchelljkotler: That is how the current form works - I would say updating this is part of the front-end changes

squarelet/organizations/forms.py
mitchelljkotler: I would also put this form change as a change to be made along with the front end

squarelet/organizations/tasks.py
mitchelljkotler: I think I was still conflating these two concepts while figuring out how to do the anchor dates.  I see how this is combining two separate ideas and it would be better if we split them apart.  This PR is already larger than I would like - I think this is a good candidate for a follow on PR.

squarelet/payments/forms.py
mitchelljkotler: I am pretty sure this got caught in an unrelated refactor - I am going to add this back in.  This change would go from allowing you to resubscribe to a cancelled but active plan, to not allowing that.

squarelet/payments/views.py
mitchelljkotler: This is only looking for subscription for the given plan, and any given organization can only have one subscription to a given plan.

duckduckgrayduck:
First test: Starting from logged out, going to /selectplan/ signing up for a new account and adding payment details for a professional plan. I reached this sentry error and it showed me a network error: 
https://muckrock.sentry.io/issues/7600304770/?alert_rule_id=472594&alert_timestamp=1783528098761&alert_type=email&environment=production&notification_uuid=3501c3a7-2c8b-41c5-bfa8-326d3b7658b5&project=1247105

mitchelljkotler:
> First test: Starting from logged out, going to /selectplan/ signing up for a new account and adding payment details for a professional plan. I reached this sentry error and it showed me a network error: https://muckrock.sentry.io/issues/7600304770/?alert_rule_id=472594&alert_timestamp=1783528098761&alert_type=email&environment=production&notification_uuid=3501c3a7-2c8b-41c5-bfa8-326d3b7658b5&project=1247105

Resolved

duckduckgrayduck:
Second path: 
Already existing user, updates their subscription to professional from their profile 
/users/test_multi1/payment/

Reaches Sentry error: https://muckrock.sentry.io/issues/7600543843/?project=1247105&query=is%3Aunresolved&referrer=issue-stream

eyeseast:
I had Claude look for landmines and got this:

> Silent billing/resource downgrade from stale org.max_users
> 
> This PR moves the subscription quantity from Organization.max_users onto Subscription.quantity, but org.max_users is no longer kept in sync, and two places still read it. The result is a bug on the single-subscription edit flow that ships today (independent of the unshipped multi-subscription UI).
> 
> Note that max_users isn't a seat cap — it's labeled "resource blocks" and is the billing quantity sent to Stripe. Nothing enforces it as a membership limit.
> 
> Failure scenario:
> 1. An org buys a plan with 10 resource blocks → sub.quantity = 10, Stripe is billed for 10, but org.max_users stays at its creation-time default (e.g. 5).
> 2. An admin reopens the Update Subscription page for any reason (even just to edit receipt emails). UpdateSubscription.get_initial() seeds the "Resource Blocks" field from self.object.max_users, so the field displays 5, not 10.
> 3. The admin saves. The form's max_users (5) flows through _apply_subscription_change → modify_subscription() → sub.quantity = 5 → Stripe modifies the subscription down to 5.
> 4. Billing and delivered resources are silently cut from 10 → 5, with no user intent and no warning. This fires on the very next edit after any purchase.
> 
> This is a regression: the old set_subscription did self.max_users = max_users; self.save(), so get_initial() always read a current value and the form round-tripped correctly. Removing that write broke the round-trip.
> 
> Secondary issue: OrganizationSerializer and fe_api/serializers.py still expose max_users. After any purchase, the API reports the stale org value (5) while the entitlements payload reports the true sub.quantity (10) — the same response is internally inconsistent, and any external consumer keying off org-level max_users sees the wrong number.
> 
> What's working correctly: purchase-time billing and the entitlement quantity delivered to clients are both driven by sub.quantity. The damage is confined to (a) the form round-trip, which then corrupts sub.quantity, and (b) the stale API field.
> 
> Suggested fix:
> - Minimal / stop the bleeding: UpdateSubscription.get_initial() should read the current subscription's quantity rather than org.max_users.
> - Cleaner: make Organization.max_users a property derived from the org's subscriptions (this fixes the serializers too), or have add_subscription/modify_subscription write quantity back to org.max_users for the interim single-subscription world.
> - Add a regression test: edit a subscription without touching the resource-blocks field and assert sub.quantity is unchanged.

eyeseast:
Testing a payment locally, I get this error:

```
squarelet_django-1        | 172.19.0.7 - - [08/Jul/2026 16:56:02] "POST /users/multiple-test/payment/ HTTP/1.1" 500 -
squarelet_django-1        | Traceback (most recent call last):
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/contrib/staticfiles/handlers.py", line 80, in __call__
squarelet_django-1        |     return self.application(environ, start_response)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/wsgi.py", line 124, in __call__
squarelet_django-1        |     response = self.get_response(request)
squarelet_django-1        |                ^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/base.py", line 140, in get_response
squarelet_django-1        |     response = self._middleware_chain(request)
squarelet_django-1        |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 57, in inner
squarelet_django-1        |     response = response_for_exception(request, exc)
squarelet_django-1        |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 141, in response_for_exception
squarelet_django-1        |     response = handle_uncaught_exception(
squarelet_django-1        |                ^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 182, in handle_uncaught_exception
squarelet_django-1        |     return debug.technical_500_response(request, *exc_info)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django_extensions/management/technical_response.py", line 41, in null_technical_500_response
squarelet_django-1        |     raise exc_value.with_traceback(tb)
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
squarelet_django-1        |     response = get_response(request)
squarelet_django-1        |                ^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
squarelet_django-1        |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
squarelet_django-1        |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/contextlib.py", line 81, in inner
squarelet_django-1        |     return func(*args, **kwds)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/views/generic/base.py", line 105, in view
squarelet_django-1        |     return self.dispatch(request, *args, **kwargs)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/contrib/auth/mixins.py", line 73, in dispatch
squarelet_django-1        |     return super().dispatch(request, *args, **kwargs)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/app/squarelet/users/views.py", line 97, in dispatch
squarelet_django-1        |     return super().dispatch(request, *args, **kwargs)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/contrib/auth/mixins.py", line 109, in dispatch
squarelet_django-1        |     return super().dispatch(request, *args, **kwargs)


squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_nginx-1         | 192.168.65.1 - - [08/Jul/2026:20:56:02 +0000] "POST /users/multiple-test/payment/ HTTP/1.1" 500 41157 "https://dev.squarelet.com/users/multiple-test/payment/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36"
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/views/generic/base.py", line 144, in dispatch
squarelet_django-1        |     return handler(request, *args, **kwargs)


squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/views/generic/edit.py", line 206, in post
squarelet_django-1        |     return super().post(request, *args, **kwargs)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/usr/local/lib/python3.12/site-packages/django/views/generic/edit.py", line 151, in post
squarelet_django-1        |     return self.form_valid(form)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/app/squarelet/organizations/views/subscription.py", line 83, in form_valid
squarelet_django-1        |     self._apply_subscription_change(organization, current_plan, user, form)
squarelet_django-1        |   File "/app/squarelet/organizations/views/subscription.py", line 68, in _apply_subscription_change
squarelet_django-1        |     organization.add_subscription(new_plan, max_users, user, token=token)
squarelet_django-1        |   File "/usr/local/lib/python3.12/contextlib.py", line 81, in inner
squarelet_django-1        |     return func(*args, **kwds)
squarelet_django-1        |            ^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/app/squarelet/organizations/models/organization.py", line 529, in add_subscription
squarelet_django-1        |     .get_current_period_end(stripe_subscription)
squarelet_django-1        |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
squarelet_django-1        |   File "/app/squarelet/organizations/payments/providers/stripe_modern.py", line 186, in get_current_period_end
squarelet_django-1        |     if items and items.data:
squarelet_django-1        |                  ^^^^^^^^^^^
squarelet_django-1        | AttributeError: 'builtin_function_or_method' object has no attribute 'data'
```

duckduckgrayduck:
@eyeseast How did you initiate the payment? Can you describe the steps taken to get to that error? I haven't encountered this specific one and would be good to capture how to replicate. Also, do you have the stripe test mode enabled locally? May be getting that locally only because it's neutered

eyeseast:
I created a new user and used one of the stripe test cards to upgrade to a professional plan, the card that's supposed to always succeed. I've definitely used it before, but it's been a while. (We should definitely document testing better.)

The UI reported a generic error: `A network error occurred. Please try again.` I copied the trace from the console.

duckduckgrayduck:
Paths I've tested in the deploy preview:

- New user, reached /selectplan/, signs up for professional plan and pays immediately -> Pass
- New user, reaches /selectplan/, creates new Organization and signs up for an organization plan -> Pass 
- New user, signs up for professional plan from their own profile -> Pass 
- New user, confirms email, creates organization, signs up for organization plan from their profile -> Pass 
- New user, confirms email, creates org and signs up for Sunlight plan (pay by card) all from /selectplan/ -> Pass
- New user, confirms email, creates org and signs up for Sunlight plan (pay by invoice) all from /selectplan/ -> Pass 


When I go to the profile page for a newly created user, and click "View all plans" I receive a 500 error page, which I think is this Sentry error: 
https://muckrock.sentry.io/issues/7600855464/?project=1247105&query=is%3Aunresolved&referrer=issue-stream

mitchelljkotler:
> I created a new user and used one of the stripe test cards to upgrade to a professional plan, the card that's supposed to always succeed. I've definitely used it before, but it's been a while. (We should definitely document testing better.)
> 
> The UI reported a generic error: `A network error occurred. Please try again.` I copied the trace from the console.

Did you rebuild the image?  This error suggest you are using an old version of the Stripe library.

mitchelljkotler:
> I had Claude look for landmines and got this:
> 
> > Silent billing/resource downgrade from stale org.max_users
> > This PR moves the subscription quantity from Organization.max_users onto Subscription.quantity, but org.max_users is no longer kept in sync, and two places still read it. The result is a bug on the single-subscription edit flow that ships today (independent of the unshipped multi-subscription UI).
> > Note that max_users isn't a seat cap — it's labeled "resource blocks" and is the billing quantity sent to Stripe. Nothing enforces it as a membership limit.
> > Failure scenario:
> > 
> > 1. An org buys a plan with 10 resource blocks → sub.quantity = 10, Stripe is billed for 10, but org.max_users stays at its creation-time default (e.g. 5).
> > 2. An admin reopens the Update Subscription page for any reason (even just to edit receipt emails). UpdateSubscription.get_initial() seeds the "Resource Blocks" field from self.object.max_users, so the field displays 5, not 10.
> > 3. The admin saves. The form's max_users (5) flows through _apply_subscription_change → modify_subscription() → sub.quantity = 5 → Stripe modifies the subscription down to 5.
> > 4. Billing and delivered resources are silently cut from 10 → 5, with no user intent and no warning. This fires on the very next edit after any purchase.
> > 
> > This is a regression: the old set_subscription did self.max_users = max_users; self.save(), so get_initial() always read a current value and the form round-tripped correctly. Removing that write broke the round-trip.
> > Secondary issue: OrganizationSerializer and fe_api/serializers.py still expose max_users. After any purchase, the API reports the stale org value (5) while the entitlements payload reports the true sub.quantity (10) — the same response is internally inconsistent, and any external consumer keying off org-level max_users sees the wrong number.
> > What's working correctly: purchase-time billing and the entitlement quantity delivered to clients are both driven by sub.quantity. The damage is confined to (a) the form round-trip, which then corrupts sub.quantity, and (b) the stale API field.
> > Suggested fix:
> > 
> > * Minimal / stop the bleeding: UpdateSubscription.get_initial() should read the current subscription's quantity rather than org.max_users.
> > * Cleaner: make Organization.max_users a property derived from the org's subscriptions (this fixes the serializers too), or have add_subscription/modify_subscription write quantity back to org.max_users for the interim single-subscription world.
> > * Add a regression test: edit a subscription without touching the resource-blocks field and assert sub.quantity is unchanged.

I will look into this one - the goal is to simply move away from max_users entirely

eyeseast:
Rebuilt the image and payment worked.

eyeseast (approved):

duckduckgrayduck (approved):

PR #693, opened 2026-06-02
OIDC Exchange Update

### LINKED ISSUE #676: Remove old JWT endpoint if unneeded
The `openid/jwt` endpoint seems unneeded and to not work how it was intended.  I added the `api/jwt` endpoint to allow clients to access the JWT tokens.  We should ensure no one is actually using `openid/jwt` and then remove it.

duckduckgrayduck:
ScopePermission in squarelet/oidc/permissions checks for auth too, which is why I didn't use it. Took a look at the Klaxon-Cloud code and it POSTs an OIDC token with no auth header. Since this is a one time use case permission, I did the check in line, but if necessary I can break it off into its own permission. 

Also removes the old JWT endpoint. This PR closes #676

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -14,7 +14,7 @@ from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
 
 # Squarelet
 from squarelet.core.views import HomeView, SelectPlanView
-from squarelet.oidc.views import OIDCRedirectURIUpdater, token_view
+from squarelet.oidc.views import OIDCRedirectURIUpdater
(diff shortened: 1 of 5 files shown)


allanlasser (approved):
Overall looks good to me. I am seeing some third-party clients in production who have `read_auth_token` within their scopes.

duckduckgrayduck:
I was told only first party clients should have read_auth_token. @mitchelljkotler can you confirm?

mitchelljkotler (approved):

PR #696, opened 2026-06-09
Fix plan purchase template to show org picker when user belongs to no orgs

### LINKED ISSUE #695: Cannot create organization from plan purchase page
When a user doesn't belong to any group organizations, they can't create one when purchasing a plan.

allanlasser:
Close #695

On Friday, we encountered an issue where a user who belongs to no orgs cannot purchase a group plan. They should be offered the option to create a new org then and there.

This also enforces the requirement that users must have a verified email before they can create an organization through the plan page. This wasn't enforced previously.

### DIFF
--- a/e2e/organizations.spec.ts
+++ b/e2e/organizations.spec.ts
@@ -117,20 +117,14 @@ test.describe("Organization Viewing", () => {
     });
 
     test("cannot see admin emails without a verified email", async ({ page }) => {
-      // Unverify e2e-regular's email so we can test the unverified case
-      runManageCommand(
-        `shell -c "from allauth.account.models import EmailAddress; EmailAddress.objects.filter(user__username='e2e-regular').update(verified=False)"`,
-      );
-
+      await login(page, "e2e-unverified");
       await page.goto("/organizations/e2e-public-org/");
       await expect(page.locator(".user .info .caption")).toHaveCount(0);
+    });
 
-      // Re-verify the email and confirm admin emails now appear
-      runManageCommand(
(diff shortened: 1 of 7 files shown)


duckduckgrayduck:
E2E fail. I'll review once they are fixed

allanlasser:
@duckduckgrayduck Passing tests now!

duckduckgrayduck:
Starting from https://squarelet-pi-fix-org-pl-uzrqke.herokuapp.com/selectplan/
I click "Premium for organizations" 
I sign up, click Continue without verifying my email. 
I click "Create new organization" with name "Test New Unverified Org"
Enter my payment details and continue. It processes my payment and succeeds. 

I think this isn't working as intended

allanlasser:
@duckduckgrayduck I get where this feedback is coming from, but I think there's two important qualifiers before considering addressing it:

1. This is in the context of the wrong PR. In this PR, we're fixing the situation where somebody isn't able to create an organization if they don't belong to an organization already. That behavior is consistent in this flow. I believe you're targeting behavior that we're addressing in #656.
2. The intention of the changes in #656 is to prevent creation of spam orgs/invitations through free organizations. In the flow you identified, the user has to make a $100 purchase to a valid credit card. I think this provides sufficient spam protection, while not interrupting the flow of somebody who wants to make a purchase. We may actually want to relax our checks so that creating an organization through a purchase is a sufficient signal of legitimacy.

Let me know if this is convincing enough to let this fix move ahead.

duckduckgrayduck (approved):

PR #709, opened 2026-06-29
Add organization data to auth serializer

### LINKED ISSUE #705: Add organization data to auth serializer
- `organization.url`
- `organization.location`

We would need to update the consent screen to indicate that this information would be shared. We would also need to update our integration guide with this information.

allanlasser:
Fixes #705

### DIFF
--- a/squarelet/organizations/serializers.py
+++ b/squarelet/organizations/serializers.py
@@ -74,6 +74,8 @@ class OrganizationDetailSerializer(OrganizationSerializer):
     update_on = serializers.SerializerMethodField()
     entitlements = serializers.SerializerMethodField()
     card = serializers.SerializerMethodField()
+    urls = serializers.SlugRelatedField(many=True, read_only=True, slug_field="url")
+    location = serializers.SerializerMethodField()
 
     class Meta:
         model = Organization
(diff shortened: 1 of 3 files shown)


squarelet/organizations/serializers.py
```diff
@@ -74,18 +74,31 @@ class OrganizationDetailSerializer(OrganizationSerializer):
...
             "card",
             "update_on",
+            "urls",
+            "location",
         )
 
     def get_update_on(self, _obj):
         return None
 
+    def get_location(self, obj):
```
mitchelljkotler, line 93: Is there a reason to return None if all are None?  This creates an extra case rather than always returning an object and have any subset of the keys possibly be None.  Not a blocker.

mitchelljkotler (approved):

squarelet/organizations/serializers.py
allanlasser: I was thinking that we'd return `null` in the response when there's no object, and that would be a more efficient check than an empty string test across all possible keys. My thinking was, "is there any location data at all? ok, now let me check which keys I can show." Especially if I was building a UI. But if you think it's better design otherwise, happy to switch this, since changes to our public API are expensive.

allanlasser: The more I think through this, the more confident I am that the two cases here are the right ergonomics. The null case is very helpful signal to ignore the field, vs. checking the entire object for if it contains any data.

PR #712, opened 2026-06-29
Handle errors thrown when creating an individual_org for a user

### LINKED ISSUE #707: Handle errors thrown when creating an individual_org for a user
Sometimes if the user was already created we get a name-collision error that's captured in Sentry; we can handle this error gracefully.

allanlasser:
Fixes #707
Fixes SQUARELET-JJ

Fixes the unhandled 500 on `/accounts/signup/` when a concurrent/double-submitted signup races past username validation and collides

## Changes
- **views.py** — `SignupView.form_valid` now catches `ValidationError` (both the standard and social branches) and re-renders the form via `form_invalid` instead of letting it bubble up as a 500.
- **managers.py** — wrapped individual-org creation in `_create_user` in a `transaction.atomic()` savepoint so a failed insert rolls back cleanly (no orphaned org, no poisoned request transaction); chained the re-raised `ValidationError` and log the collision as a warning.
- **serializers.py** — `UserWriteSerializer.create` converts the Django `ValidationError` to a DRF `ValidationError` so the API path returns 400, not 500.

## Tests
- `test_managers.py` — duplicate username (incl. case-insensitive) raises a clean `ValidationError` with no orphaned org.
- `test_views.py` — `SignupView.form_valid` handles the collision gracefully on both the standard and social branches.

### DIFF
--- a/squarelet/users/managers.py
+++ b/squarelet/users/managers.py
@@ -3,10 +3,17 @@ from django.contrib.auth.models import UserManager as AuthUserManager
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError, transaction
 from django.db.models import Q
+from django.utils.translation import gettext_lazy as _
+
+# Standard Library
+import logging
+import sys
 
 # Squarelet
 from squarelet.core.utils import mailchimp_journey
 from squarelet.organizations.models import Organization
+
+logger = logging.getLogger(__name__)
 
 
 class UserManager(AuthUserManager):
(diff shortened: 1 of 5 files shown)


duckduckgrayduck:
I was finally able to replicate this (and see it) by using a Playwright script to click the Sign up button in two different browsers at the same time, resulting in the Sentry error:
[](https://muckrock.sentry.io/issues/7583460573/?project=1247105&query=is%3Aunresolved&referrer=issue-stream)

When I do the same with the script pointed at this deploy preview, it no longer throws the error and it instead correctly bubbles up the message to the failing browser. 
A user with that username already exists. Adding Mitch for code review

mitchelljkotler (approved):

PR #716, opened 2026-07-06
Redesign Squarelet billing pages

### LINKED ISSUE #674: Showing incorrect receipts
When a staff user clicks on "View Receipts" for an org they are taking a look at an issue for, it shows the staff user's current orgs' receipts instead of that org

### LINKED ISSUE #677: Redesign subscription page
The subscription page needs to be redesigned to be brought into line with the rest of the redesigned pages.

We will also need to support being able to subscribe to and manage multiple subscriptions.

We also want to ensure we are following best practices in regards to making it clear how to unsubscribe:
https://www.notion.so/muckrock/Review-ease-of-cancellation-31ef8892696380ffbe07decb93f30df7

dnass:


squarelet/organizations/views/subscription.py
```diff
@@ -150,6 +201,126 @@ def get_initial(self):
         }
 
 
+class UpdateCard(OrganizationPermissionMixin, UpdateView):
```
dnass: Is `OrganizationPermissionMixin` appropriate to use here and on the other new view classes?

squarelet/organizations/urls.py
```diff
@@ -15,9 +15,31 @@
...
+    path(
+        "<slug:slug>/subscriptions/",
+        view=views.ManageSubscriptions.as_view(),
+        name="subscriptions",
+    ),
     path(
         "<slug:slug>/payment/", view=views.UpdateSubscription.as_view(), name="payment"
     ),
+    path(
+        "~cancel/<int:pk>/",
```
dnass: Am I right to understand that paths start with `~` when the corresponding view is querying something other than an `Organization`, so the URL captures the pk rather than the org slug?

squarelet/organizations/views/subscription.py
```diff
@@ -150,6 +201,126 @@ def get_initial(self):
...
+
+    def get_initial(self):
+        return {
+            "receipt_emails": ", ".join(
+                r.email for r in self.object.receipt_emails.all()
+            ),
+        }
+
+
+class CancelSubscription(OrganizationPermissionMixin, DeleteView):
```
dnass: Not sure if `DeleteView` is actually the right base class here

squarelet/organizations/views/subscription.py
```diff
@@ -150,6 +201,126 @@ def get_initial(self):
...
+
+        subscription = self.object.subscriptions.first()
+        if subscription:
+            context["organization"] = subscription.organization
+            context["next_date"] = get_subscription_next_date(subscription)
+
+        return context
+
+
+class PaymentsList(ListView):
```
dnass: When I include `OrganizationPermissionMixin` here, I get `AttributeError: 'PaymentsList' object has no attribute 'get_object'`. I believe that's because a `ListView` has an `object_list` rather than a single object. Not sure if the permission mixin is even needed here or if there's another approach that would work.

squarelet/templates/subscriptions/manage_subscriptions.html
```diff
@@ -0,0 +1,179 @@
...
+  <section id="plans">
+    <header>
+      <div class="title">
+        <h2>{% trans "Your plans" %}</h2>
+      </div>
+    </header>
+
+    {% for subscription in subscriptions %}
+      {% with annual=subscription.plan.annual %}
+        <main id="{{ subscription.plan.slug }}" class="subscription-card">
```
dnass, line 72: Is it safe to use the plan slug as the ID here, or could an org ever have multiple subscriptions to the same plan?

squarelet/organizations/views/subscription.py
allanlasser: Yep, `OrganizationPermissionsMixin` should be used when we want to restrict some views to org admins (plus MuckRock staff)—see `docs/contributors/permissions.md` for more detail on how permissions work.

The `can_edit_subscription` rule is set in `squarelet/organization/rules/organizations.py`.

squarelet/organizations/views/subscription.py
allanlasser: Good flag—I don't think this is the right base class. When cancelled, a subscription shoudl stick around until the end of its billing cycle. This is _especially_ true for annual subscriptions, where the subscription may remain active for months post-cancellation.

Instead, we should call `organization.remove_subscription(subscription)` when the form is valid and we should "cancel" the sub.

Any subscriptions with `cancelled = True` will be automatically deleted the next time the `restore_organization` task runs.

While reviewing this, I noticed an existing bug in our logic for cleaning up canceled subscriptions. Filed as #717.

squarelet/organizations/views/subscription.py
allanlasser: Oh, interesting. In this case, I think the permission mixin's check wouldn't work, since the object returned by `get_object` is not an `Organization`. Instead, looking at [how the `OrganizationPermissionMixin` is defined](https://github.com/MuckRock/squarelet/blob/9184095cfeef33471949ee2fb6b3c029b4d5c2ea/squarelet/organizations/mixins.py#L16-L17), I think the right approach would be:

1. Extend `PermissionRequiredMixin` in this class
2. Override `has_permission` to check `user.has_perm("organizations.can_view_charge", organization)`

squarelet/organizations/urls.py
allanlasser: After checking with Mitch, I confirmed that these paths are a holdover from a previous (now-unused) templating library. I think they are intended to provide a single, stable URL for viewing any single subscription, with the organization coming through the relation. That being said, I'm not thrilled with this URL scheme.

Now that we have multi-subscription support, I think it'd be better to follow the pattern set by `<slug:slug>/subscriptions/` and go with something like `<slug:slug>/subscriptions/<int:pk>/cancel`, so that each URL contains both an org slug and a subscription ID.

squarelet/templates/subscriptions/manage_subscriptions.html
allanlasser: We're enforcing the rule that orgs can only ever have 1 subscription to a plan. We're safe to use the slug as an ID here!

squarelet/organizations/views/subscription.py
dnass: I did use `OrganizationPermissionMixin` with other views that don't query `Organization`s (`UpdateSubscriptionFrequency` and `CancelSubscription` both modify `Plan`s). Those ones don't throw errors but maybe they're failing in a more subtle way?

squarelet/organizations/urls.py
dnass: Sounds good

squarelet/organizations/views/subscription.py
dnass: I added `PermissionRequiredMixin` to `PaymentsList`. 

I also confirmed that in [`UpdateSubscriptionFrequency`](https://github.com/MuckRock/squarelet/blob/feae7ebe97e976ec52bd1cbecad0db2532380a2e/squarelet/organizations/views/subscription.py#L249-L252), which queries subscriptions, `OrganizationPermissionMixin` is checking the user permissions against a `Subscription` object [here](https://github.com/MuckRock/squarelet/blob/9184095cfeef33471949ee2fb6b3c029b4d5c2ea/squarelet/organizations/mixins.py#L25). I can refactor the view class to query the org instead, but I wonder if `OrganizationPermissionMixin` should also safeguard against misuse — should `has_permission` return false or throw an error if `obj` isn't an `Organization`?

squarelet/organizations/views/detail.py
```diff
@@ -52,22 +49,18 @@ def get_context_data(self, **kwargs):
             context["users"] = admins
         context["admins"] = admins
 
-        # Get the current plan and subscription, if any
-        current_plan = None
+        # Get subscriptions, if any
         upgrade_plan = Plan.objects.get(slug="organization")
-        subscription = None
+        subscriptions = None
         if hasattr(org, "subscriptions"):
```
mitchelljkotler: I am not sure why this check is here, org will always have the `subscriptions` attribute.  This can just be:
`context["subscriptions"] = org.subscriptions.all()`

squarelet/subscriptions/views.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Get subscriptions and add renewal/cancellation date and cost data
+        subscriptions = self.object.subscriptions.all()
+        for subscription in subscriptions:
+            subscription.next_date = get_subscription_next_date(subscription)
+            subscription.cost = subscription.plan.base_price
+        context["subscriptions"] = subscriptions
+
+        # Get card on file
+        customer = self.object.customer()
```
mitchelljkotler: Having to write this check out every time is not good - I left this in my branch, but I am going to update this so  customer.card just returns the actual card object you want - I will let you know when that is merged in so you can update on your branch

squarelet/subscriptions/views.py
```diff
@@ -0,0 +1,205 @@
...
+        context["payments"] = payments
+
+        return context
+
+
+class BaseUpdateSubscriptionFrequency(UpdateView):
+    form_class = UpdateSubscriptionFrequencyForm
+    template_name = "subscriptions/update_subscription_frequency.html"
+
+    def get_queryset(self):
```
mitchelljkotler: You shouldn't need to override get_queryset or get_object here - it should do the right thing if you pass in a slug as a kwarg from the URL

squarelet/subscriptions/views.py
```diff
@@ -0,0 +1,205 @@
...
+                r.email for r in self.object.receipt_emails.all()
+            ),
+        }
+
+
+class BaseCancelSubscription(UpdateView):
+    form_class = CancelSubscriptionForm
+    template_name = "subscriptions/cancel_subscription.html"
+
+    def get_queryset(self):
```
mitchelljkotler: Same here

squarelet/users/views.py
```diff
@@ -159,35 +164,23 @@ def get_context_data(self, **kwargs):
         context["unused_code_count"] = len(self.get_recovery_codes())
         # Get the current plan and subscription, if any
         individual_org = user.individual_organization
-        current_plan = None
         upgrade_plan = Plan.objects.filter(slug="professional").first()
-        subscription = None
+        subscriptions = None
         if hasattr(individual_org, "subscriptions"):
```
mitchelljkotler: can skip this hasattr check

mitchelljkotler (commented):
This looks like it is still under development, let me know if you have any questions

squarelet/subscriptions/views.py
dnass: The route 404s when I remove these overrides. The URL also contains the subscription pk, could that be the issue?

dnass:
@mitchelljkotler Thanks for reviewing. The only piece of this I haven't tackled yet is switching from a monthly plan to annual or vice versa. I'm happy to take a crack at it if you can lay out the general approach for me.

allanlasser:
I made a few fixes and one change so far:

1. FIX: calls `individual_organization.subscriptions.first()` when constructing context for the `/selectplan` view. It was erroring since it hadn't been updated to the multi-subscription assumption.
2. FIX: adds proper card-call mocking for tests
3. CHANGE: prefers `username` as the URL param for individual organizations, not the slug of the individual_organization. The slug and the username _will_ diverge, like for my username (`lasser.allan` gets slugified to `lasserallan`), which breaks some pages. Individual organizations are an invisible concept for users, so it's better to prefer the username here. I added a class mixin for finding the right object based on whether we're dealing with an individual or group org.

Flagging that this should be rebased against `master`, and there'll be some merge conflicts along the way.

allanlasser (commented):
We already have a `squarelet/payments/forms.py` and `squarelet/payments/views.py`—let's update those files intead of adding the `subscriptions` module.

squarelet/organizations/views/subscription.py
mitchelljkotler: I think throwing an error makes sense - it is a misconfiguration, I think you can just add an `assert isinstance(obj, Organization)` into the `has_permission` method to prevent misuse

dnass: Great, just made that change

allanlasser (commented):
- [x] Need a consolidated view of all payments _in addition to_ a payment history for a specific org. Let's use the templates/subscriptions/payments.html as a base for two layouts:
  1. A template "organization_payments.html" that lists just payments for a single org.
  2. A template "payments_hub.html" that list all payments a user can see, across all orgs they belong to. Just show the 5 most recent payments, with a link to view all payments for the org—this leads to the organziation_payments template.
- [x] Support removing cards
- [x] Activity stream logging when staff take actions for accountability
- [x] Update "Upgrade" button link (currently points to old payment page)
- [x] ~Modify current subscription in-place (upgrade/downgrade) for pro-ration benefits. No support for adding second subscription yet—will come in follow-up change.~ Moving this handling to a new branch and PR
- [x] ~~Changing plan frequency is unimplemented~~ Move change frequency feature to a new branch and PR #755 
- [x] #753

allanlasser:
Screenshot 2026-07-24 at 09 36 11

Belonging to an organization with multiple subscriptions creates some unintentional chaos on user profile pages. This worked well when there was just 1 subscription. Perhaps we can consolidate this into a single card—the most important information is the benefits a user inherits, and making that clear to them.

**Update:** I'm thinking about tying the benefits list data to entitlements; this way, benefits are an "plain language" description of the entitlement and they can be composed on plans.

**Update 2:** We can handle this separately without blocking this branch. I opened #753 to address.

squarelet/templates/organizations/includes/plan_card.html
```diff
@@ -17,20 +17,27 @@
...
-        <div class="plan-detail">
-          <h3>{% trans "Payment card" %}</h3>
-          <span class="caption">
-            <span class="icon">{% icon "credit-card" %}</span>
-            <span class="text">
-              {{ card_brand }} {% trans "ending in" %} <strong>{{ card_last4 }}</strong>
+      <div class="account-row">
+        <div class="account-card">
+          <span class="icon">{% icon "credit-card" %}</span>
+          {% if account.card_brand and account.card_last4 %}
```
dnass: @allanlasser The `account` object doesn't exist in this context, so this section always shows "no card on file"

allanlasser: Oops, good catch. Bad copy/paste from another view.

allanlasser:
We're going to have to punt on the frequency change—we don't have this set up in Stripe or in our backend. I'll fork this out this UI into a separate branch as the starting point for the backend work.

allanlasser:
For modifying vs. adding, we will just handle modifying in this PR. Then, we'll follow-up on a new focused PR that supports modifying OR adding a subscription.

allanlasser:
This code is in great shape. I've moved any further changes onto dedicated branches for smaller, interative PRs upon this foundation. I'm merging this into a release-ready branch so we can mark it done, but hold off on merging until all the follow-on work is complete.

PR #728, opened 2026-07-10
Daily odoo sync

duckduckgrayduck:
All methods and logic have been tested in the shell (dry run and live) and used to maintain the sync between Accounts and Odoo. This would just schedule it.

### DIFF
--- a/squarelet/core/management/commands/sync_odoo.py
+++ b/squarelet/core/management/commands/sync_odoo.py
@@ -1,0 +1,757 @@
+# Django
+from django.conf import settings
+from django.core.mail import EmailMessage
+from django.core.management.base import BaseCommand
+
+# Standard Library
+import logging
+from datetime import date
+from io import StringIO
+
+# Third Party
+import requests
+
+# Squarelet
+from squarelet.core.utils import requests_retry_session
+from squarelet.organizations.models.organization import Membership, Organization
+from squarelet.organizations.models.payment import Plan
+
+logger = logging.getLogger(__name__)
+
+SKIP_SLUGS = {"sunlight-search"}
+
+# Collaborative config: slug -> (tag_id, label)
+COLLABORATIVES = {
+    "rural-news-network": (1, "RNN"),
+    "granite-state-news-collaborative-fiscally-sponsore": (3, "GSNC"),
+}
+
+# Cache of x_plan name -> id
+_PLAN_ID_CACHE = {}
+
+# Default page size for paginated search_read calls
+_PAGE_SIZE = 200
+
+
+def _headers():
+    return {
+        "Authorization": f"bearer {settings.ODOO_API_KEY}",
+        "Content-Type": "application/json",
+    }
+
+
+_session = requests_retry_session()
+
+
+def _odoo_request(endpoint, payload):
+    """POST to an Odoo JSON-2 endpoint via the retry session.
+    Uses requests_retry_session which will retry on intermittent issues.
+    Returns parsed JSON on success, or raises on failures that are ongoing.
+    The exception propagates to handle, which catches it,
+    sets failed = True, and emails the failure report."""
+    url = f"{settings.ODOO_URL}/json/2/{endpoint}"
+    try:
+        resp = _session.post(
+            url,
+            headers=_headers(),
+            json=payload,
+            timeout=settings.ODOO_TIMEOUT,
+        )
+        resp.raise_for_status()
+        return resp.json()
+    except requests.exceptions.RequestException as exc:
+        logger.error("Odoo request failed for %s: %s", endpoint, exc)
+        raise
+
+
+def odoo_search(model, domain, fields, limit=1, offset=0):
+    result = _odoo_request(
+        f"{model}/search_read",
+        {"domain": domain, "fields": fields, "limit": limit, "offset": offset},
+    )
+    return result if result is not None else []
+
+
+def odoo_search_all(model, domain, fields, page_size=_PAGE_SIZE):
+    """search_read every matching row, paging so nothing is capped.
+    Stops when a batch comes back smaller than page_size."""
+    results = []
+    offset = 0
+    while True:
+        batch = odoo_search(model, domain, fields, limit=page_size, offset=offset)
+        results.extend(batch)
+        if len(batch) < page_size:
+            break
+        offset += page_size
+    return results
+
+
+def odoo_create(model, vals):
+    return _odoo_request(f"{model}/create", {"vals_list": [vals]})
+
+
+def odoo_write(model, ids, vals):
+    return _odoo_request(f"{model}/write", {"ids": ids, "vals": vals})
+
+
+def _resolve_plan_id(name):
+    """Resolve a Squarelet plan to its Odoo x_plan id by name match.
+    Creation is handled up front by _ensure_all_plans, so this is
+    lookup-only; a missing plan means it wasn't in Squarelet at ensure time."""
+    if name in _PLAN_ID_CACHE:
+        return _PLAN_ID_CACHE[name]
+    res = odoo_search("x_plan", [["x_name", "=", name]], ["id"])
+    pid = res[0]["id"] if res else None
+    if pid is None:
+        logger.warning("No Odoo plan match for Squarelet plan: %s", name)
+    _PLAN_ID_CACHE[name] = pid
+    return pid
+
+
+def _build_plan_vals(plan):
+    """Map a Squarelet Plan to the x_plan fields Odoo mirrors."""
+    return {
+        "x_name": plan.name,
+        "x_studio_slug": plan.slug,
+        "x_studio_base_price": plan.base_price,
+        "x_studio_price_per_user": plan.price_per_user,
+        "x_studio_annual": plan.annual,
+    }
+
+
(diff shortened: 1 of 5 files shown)


squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+
+# Third Party
+import requests
+
+# Squarelet
+from squarelet.organizations.models.organization import Membership, Organization
+
+logger = logging.getLogger(__name__)
+
+ODOO_URL = "https://muckrock-odoo.odoo.com"
```
mitchelljkotler: This URL should probably be configurable

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+ODOO_API_KEY = settings.ODOO_API_KEY
```
mitchelljkotler: Defining this constant here locks in the variable value at import time - I think this will work on Heroku, since it restarts if you update env vars, but is probably generally safer if we do it at run time.  Just use `settings.ODOO_API_KEY` directly.  `HEADERS` should also be a function that returns that dictionary so it isn't set at import time.

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+
+HEADERS = {
+    "Authorization": f"bearer {ODOO_API_KEY}",
+    "Content-Type": "application/json",
+}
+
+ODOO_TIMEOUT = 30
```
mitchelljkotler: Might as well make this configurable as well

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+        "x_studio_slug": org.slug,
+        "x_studio_muckrock_accounts_id": org.id,
+        "x_studio_muckrock_accounts_uuid": str(org.uuid),
+        "x_studio_muckrock_accounts": True,
+        "city": org.city or "",
+        "website": ", ".join(urls),
+        "x_studio_verified_journalist": org.verified_journalist,
+        "is_company": True,
+        "company_type": "company",
+        "x_studio_plan_1": [(6, 0, odoo_plan_ids)],
```
mitchelljkotler, line 159: Could use a comment here explaining the format - what do the 6 and 0 represent?

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+ODOO_API_KEY = settings.ODOO_API_KEY
+
+HEADERS = {
+    "Authorization": f"bearer {ODOO_API_KEY}",
+    "Content-Type": "application/json",
+}
+
+ODOO_TIMEOUT = 30
+
+PLAN_SLUG_MAP = {
```
mitchelljkotler: Hard coding IDs can be problematic - these should be loaded from the database.  Different environments could have different IDs, and changing plans should not cause this code to need to change.

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+    return pid
+
+
+def _compute_org_plans_and_status(org, inherited_plan_ids):
+    """Return (odoo_plan_ids, sunlight_status) for an org."""
+    plans = list(org.plans.values_list("name", "slug"))
+    sunlight_plans = [
+        name
+        for name, slug in plans
+        if "sunlight" in name.lower()
```
mitchelljkotler: We should move away from slug comparisons - we might need a boolean on plans to specify if they are sunlight plans or not

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+            logger.info("[DRY RUN] Would create org: %s (slug=%s)", org.name, org.slug)
+            return None, odoo_plan_ids
+        else:
+            result = odoo_create("res.partner", vals)
+            odoo_id = result[0]
+            logger.info("Created org: %s", org.name)
+            return odoo_id, odoo_plan_ids
+
+
+# pylint:disable=too-many-locals
```
mitchelljkotler: I'd prefer you refactor this down into multiple functions rather than suppressing the error - it is easier to reason about smaller functions

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+    "rural-news-network": (1, "RNN"),
+    "granite-state-news-collaborative-fiscally-sponsore": (3, "GSNC"),
+}
+
+# Cache of x_plan name -> id for plans resolved outside PLAN_SLUG_MAP
+_PLAN_ID_CACHE = {}
+
+
+def odoo_search(model, domain, fields, limit=1):
+    resp = requests.post(
```
mitchelljkotler: These helper functions should all check the response code before calling `resp.json` for proper error handling.  Can probably do a few retries in case it is an intermittent error, and otherwise log and move on, so one badly formatted piece of data doesn't break the whole sync.

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+
+            _sweep_lapsed_orgs(active_slugs, dry_run=dry_run, only_slug=slug)
+            for config in collaborative_data.values():
+                if config.member_slugs:
+                    _sweep_stale_collaborative_tags(
+                        config, dry_run=dry_run, only_slug=slug
+                    )
+
+            logger.info("Sync complete")
+        finally:
```
mitchelljkotler, line 752: Finally block means the sync report will always go out, but you do not catch exceptions, so there will be no indication that it failed.  You should catch exceptions and include the exception detail if it exists.

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+        [["parent_id", "=", odoo_org_id], ["is_company", "=", False]],
+        [
+            "id",
+            "email",
+            "x_studio_muckrock_accounts",
+            "x_studio_org_departed_date",
+            "x_studio_secondary_email",
+            "x_studio_plan_1",
+        ],
+        limit=200,
```
mitchelljkotler: Can this be paginated rather than limited to 200?  What happens if there are more than 200 members?

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,685 @@
...
+            "x_studio_muckrock_accounts_uuid": current[
+                "x_studio_muckrock_accounts_uuid"
+            ],
+        }
+        diffs = {
+            k: (normalized[k], v) for k, v in vals.items() if normalized.get(k) != v
+        }
+        if "parent_id" in diffs and normalized.get("parent_id") is not None:
+            logger.info(
+                "Note: %s already linked to parent ID %s, skipping reassignment to %s",
```
mitchelljkotler: Is which parent wins arbitrary?

mitchelljkotler (changes requested):
Looks very good overall.  A few changes requested.  It would probably also be good to include some unit tests for this.

squarelet/core/management/commands/sync_odoo.py
duckduckgrayduck: It's the parent that was first set on Odoo. This currently only affects Amanda's account and I don't want it resetting her parent to BU Spark basically, because she is actually a member of the MuckRock org. It would be good to have some way to set "primary org" on Odoo so that it doesn't keep trying this

squarelet/core/management/commands/sync_odoo.py
duckduckgrayduck: Wasn't a case I ran into yet (an org having more than 200 members), but that's the limit via the API per call. I've added pagination

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,757 @@
...
+from squarelet.core.utils import requests_retry_session
+from squarelet.organizations.models.organization import Membership, Organization
+from squarelet.organizations.models.payment import Plan
+
+logger = logging.getLogger(__name__)
+
+SKIP_SLUGS = {"sunlight-search"}
+
+# Collaborative config: slug -> (tag_id, label)
+COLLABORATIVES = {
```
mitchelljkotler: These should be queried from the db using the `collective_enabled` field

squarelet/core/management/commands/sync_odoo.py
```diff
@@ -0,0 +1,757 @@
...
+        "x_studio_annual": plan.annual,
+    }
+
+
+def _ensure_all_plans(dry_run=False):
+    """Create an Odoo x_plan with full pricing data for any Squarelet
+    plan that lacks one. Runs before sync so every plan resolves to a
+    real id and no plan silently drops out of an org's plan list."""
+    for plan in Plan.objects.all():
+        if plan.name in _PLAN_ID_CACHE:
```
mitchelljkotler: Would this be clearer using `elif`s instead of `continue`s?

mitchelljkotler (changes requested):

mitchelljkotler (approved):

PR #732, opened 2026-07-14
Adds option to subscriptions to _not_ automatically renew

### LINKED ISSUE #730: Add option to plans to not automatically renew
Automatic renewal is enabled for all plans at time of purchase. Some plans, like high-value annual plans, should not automatically renew.

allanlasser:
Fixes #730

- Adds `auto_renew` boolean to Plans, defaults to `True`
- Updates our `Subscription.create` and `Subscription.modify` to use Stripe's `cancel_at_period_end` option, which passes-through the plan's auto-renew setting


mitchelljkotler (approved):
Looks good to me.  My one concern is what is our plan for un-cancelling the plan if they do want to renew?

allanlasser:
That's a good question. If the subscription is active and cancelled, we can provide a way for the user to opt-in to renewal. This would unset the `cancelled` flag and update the `cancel_at_period_end` to `False`. We could encourage users to provide a payment method (CC or bank account) when setting up renewal. I think this makes sense as a standalone follow up to #677.

Issue #719, opened 2026-07-07
Disentangle payment date and resource update date for organization's subscriptions


mitchelljkotler:
These both currently use `update_on` but they should be separated to support plans which update their resources on a different cadence then their payment.

PR #735, opened 2026-07-14
separate out anchor date from update on

### LINKED ISSUE #717: Subscription cleanup logic will delete active annual plans
When someone chooses to cancel their annual plan, it should stay active until the end of the that payment period. Our current cleanup logic assumes all plans are monthly, so a non-renewing annual plan will be deleted too soon.

https://github.com/MuckRock/squarelet/blob/9184095cfeef33471949ee2fb6b3c029b4d5c2ea/squarelet/organizations/tasks.py#L51-L54

Annual plans are our highest-value plans, so this fix is urgently needed before it negatively impacts a customer.

mitchelljkotler:
This fixes issue #717


squarelet/organizations/tasks.py
```diff
@@ -48,10 +48,12 @@ def restore_organization():
...
 
-    # Delete cancelled subscriptions for due orgs
+    # Delete cancelled subscriptions for due orgs where the Stripe cancellation
+    # date has passed (or is null, which covers free plans and legacy records).
     Subscription.objects.filter(
-        organization_id__in=due_org_ids, cancelled=True
-    ).delete()
+        organization_id__in=due_org_ids,
+        cancelled=True,
+    ).filter(Q(cancel_at__lte=today) | Q(cancel_at__isnull=True)).delete()
```
allanlasser, line 56: Confirming that this is the line that fixes #717 

Overall, it seems like this branch is actually fixing #719

squarelet/organizations/migrations/0069_organization_billing_anchor.py
```diff
@@ -0,0 +1,26 @@
+# Django
+from django.db import migrations, models
+from django.utils.translation import gettext_lazy as _
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("organizations", "0068_subscription_quantity"),
```
allanlasser, line 9: This migration will conflict with #732, which also adds a 0069 migration to organizations.

squarelet/organizations/tasks.py
mitchelljkotler: This ensures subscriptions are only deleted on their cancellation date.  The `cancel` method in `Subscription` sets the cancellation date when cancel is called.  Separating out `billing_anchor` is setting up for #719 but the changes here are mostly around cancelling.

squarelet/organizations/models/payment.py
```diff
@@ -446,6 +463,7 @@ def stripe_modify(self):
                 days_until_due=30 if self.plan.annual else None,
             )
             self.cancelled = False
+            self.cancel_at = None
```
allanlasser: This will conflict with #732 since it assumes the modified subscription will always renew.

allanlasser: `cancel_at` will also need to be updated when creating a subscription with `auto_renew=False`

allanlasser (commented):
I noticed some conflicts with #732—what order would you like to address these?

squarelet/organizations/migrations/0069_organization_billing_anchor.py
mitchelljkotler: I rebased on top of #732 - required a bit of code, so good to review again

allanlasser (approved):

PR #737, opened 2026-07-16
Show verification notice during OIDC authorization

allanlasser:
AgendaWatch limits access to verified users, but there were some issues with how we notify users that they need verification to access the resource.

I added a field to `ClientProfile` to indicate whether the app relies on verification in some way.

If an OIDC client relies on verification, we check the user's verified status during OIDC authorization. We shows an interstitial notice about verification, with a customizable Markdown message, and provide links for the user to request verification for one of their orgs. If they don't belong to any orgs, we give them a link to request individual verification. If a user doesn't have a confirmed email, they can't request verification.

The fact we showed this message is saved to session storage, so we only show once per client per signed-in session.

This change probably means we can remove the verification check from the onboarding flow—it’s a much more robust solution.

### Unverified, with orgs
Screenshot 2026-07-16 at 14 40 31

### Unverified, without orgs
Screenshot 2026-07-16 at 14 39 33

### Unconfirmed email
Screenshot 2026-07-16 at 14 48 22


duckduckgrayduck (approved):
Nice work!

PR #738, opened 2026-07-16
UIs for managing member orgs and invitations

### LINKED ISSUE #614: UI for enterprise orgs to control which orgs are members
We have handlers for an admin of an org with collective sharing to add and remove other organizations as children/members, but no UI for this yet. Adding self-service support for enterprise orgs will make it easier to onboard and manage these complex, multi-organization groups.

dnass:
Closes #614 

- Add Groups and Member Orgs sections to org detail view, with actions for accepting or declining group invitations.
- Create member org management view for creating, resending, and withdrawing org invitations.
- Create org invite email template.
- Add `organizations.can_manage_member_orgs` permission.
- Extract styled Svelecte into a reusable select component.

image
image
image
image


frontend/components/SelectChip.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+  import type { Snippet } from "svelte";
+
+  let { type, content }: { type?: string; content: Snippet } = $props();
+</script>
+
+<div class={["chip", { email: type === "email" }]}>
+  {@render content?.()}
```
allanlasser: We can improve this component in 1 of 2 ways:

1. use Svelte's [implicit children snippet](https://svelte.dev/docs/svelte/snippet#Passing-snippets-to-components-Implicit-children-snippet) instead of `content`, or
2. render the "deselect" button in here, inside a "deselect" snippet.

Either approach improves reuse. You can't use children when the component has other content, so it's an either-or choice.

If we go with 1, a "deselect button" component would help maintain consistency across instances.

squarelet/organizations/views/detail.py
```diff
@@ -136,6 +139,34 @@ def _get_subscription_context(self, org, subscription):
...
+            )
+
+        ctx["show_groups_section"] = is_admin and (
+            len(ctx["groups"]) > 0 or len(ctx["group_invitations"]) > 0
+        )
+
+        # For collective groups, the members section is always visible to admins,
+        # and visible to non-admins only when the group has member organizations.
+        ctx["show_members_section"] = org.collective_enabled and (
+            is_admin or (is_member and len(ctx["members"]) > 0)
```
allanlasser: Instead of checking on `is_admin`, check on `user.has_perm("organization.can_manage_member_orgs")`. This will include admins along with any MuckRock staff who are manually granted the permission.

squarelet/organizations/views/detail.py
```diff
@@ -136,6 +139,34 @@ def _get_subscription_context(self, org, subscription):
...
+        }
+
+        if is_admin:
+            ctx["group_invitations"] = (
+                OrganizationInvitation.objects.pending()
+                .filter(to_organization=org)
+                .select_related("from_organization")
+            )
+
+        ctx["show_groups_section"] = is_admin and (
```
allanlasser: Our support team should be able to help admins manage their orgs, and they should have permission to see the groups section. I _think_ we'd be well served here with a "can_accept_org_invites" (or similar) permission. Then, this permission can be manually granted to staff members.

squarelet/organizations/views/detail.py
```diff
@@ -315,6 +346,117 @@ def handle_sync_wix(self, request):
...
+            return
+
+        is_group_admin = org.has_admin(self.request.user)
+        is_member_admin = member_org.has_admin(self.request.user)
+
+        if not is_group_admin and not is_member_admin:
+            messages.error(
+                request, _("You do not have permission to remove this membership")
+            )
+            return
```
allanlasser: Use permissions checks here, not admin role—staff with permission should be permitted to take this action.

allanlasser (changes requested):
I'm tweaking the frontend templates a bit, in the meantime I have some actionable feedback:

squarelet/organizations/views/detail.py
dnass: Good call. Maybe `can_manage_groups` so that it encompasses accepting invites and leaving groups?

frontend/components/SelectChip.svelte
dnass: I took a slightly different route: moved the chip styles into the Select component's [default `selection` snippet](https://github.com/MuckRock/squarelet/blob/7b186a96e99739796092986f97863c58cdec3f78/frontend/components/Select.svelte#L10-L26), and introduced a new `selectValue` snippet for just the chip contents. `selection` can still be overridden if there's ever a need for a different approach in the future.

allanlasser (approved):
I went through and made some frontend changes—I think this is ready to go!

PR #744, opened 2026-07-21
fix anchor date

mitchelljkotler:


duckduckgrayduck:
I believe this would fix the issue. I'd like to flag that Stripe also has month-aware time differences built in as well:
https://docs.stripe.com/billing/subscriptions/billing-cycle#use-billing_cycle_anchor_config

duckduckgrayduck (approved):

mitchelljkotler:
> I believe this would fix the issue. I'd like to flag that Stripe also has month-aware time differences built in as well: https://docs.stripe.com/billing/subscriptions/billing-cycle#use-billing_cycle_anchor_config

I did not know this was a thing - this simplifies things greatly, I will use this instead

PR #747, opened 2026-07-22
Separate out payment methods

### LINKED ISSUE #741: Separate out Payment Method details into its own model
The data model PR added field for payment methods directly to the Customer.  Allan pointed out that this would be useful to break into its own model, for both clarity and allow us to support multiple payment methods on file in the future if we wanted to.  I did not want to increase the size of that PR, so will implement this separately.

mitchelljkotler:
This pulls payment methods (credit cards and bank accounts) out into their own model.  This helps organize the data and opens the path to supporting multiple payment methods in the future.


squarelet/organizations/models/payment.py
```diff
@@ -173,30 +167,101 @@ def payment_details(self):
...
+    @property
+    def payment_exp_year(self):
+        pm = self.default_payment_method_obj()
+        return pm.exp_year if pm else None
+
+    @property
+    def stripe_payment_method_id(self):
+        pm = self.default_payment_method_obj()
+        return pm.stripe_id if pm else ""
+
```
allanlasser, line 216: I like how you preserve the current API here!

squarelet/organizations/models/payment.py
```diff
@@ -1095,3 +1155,47 @@ class Meta:
...
+                name="unique_default_per_customer",
+            ),
+        ]
+
+    def __str__(self):
+        return f"{self.get_method_type_display()}" f" {self.brand} x{self.last4}"
+
+    @property
+    def display(self):
+        if self.brand and self.last4:
```
allanlasser, line 1199: Will `brand` be available on bank accounts, too? If not, let's ensure we support a fallback when brand's not available.

allanlasser (changes requested):
There are 4 lifecycle events from Stripe that we should handle for payment methods, if we aren't already:

- [`payment_method.attached`](https://docs.stripe.com/api/events/types#event_types-payment_method.attached)
- [`payment_method.automatically_updated`](https://docs.stripe.com/api/events/types#event_types-payment_method.automatically_updated)
- [`payment_method.detached`](https://docs.stripe.com/api/events/types#event_types-payment_method.detached)
- [`payment_method.updated`](https://docs.stripe.com/api/events/types#event_types-payment_method.updated)

allanlasser:
We don't think we need any new UI, on top of what we've done in #716. We just need to update templates/context in this PR to make sure the data flows through the same as before.

mitchelljkotler:
> There are 4 lifecycle events from Stripe that we should handle for payment methods, if we aren't already:
> 
>     * [`payment_method.attached`](https://docs.stripe.com/api/events/types#event_types-payment_method.attached)
> 
>     * [`payment_method.automatically_updated`](https://docs.stripe.com/api/events/types#event_types-payment_method.automatically_updated)
> 
>     * [`payment_method.detached`](https://docs.stripe.com/api/events/types#event_types-payment_method.detached)
> 
>     * [`payment_method.updated`](https://docs.stripe.com/api/events/types#event_types-payment_method.updated)

These are being handled now

squarelet/organizations/models/payment.py
mitchelljkotler: The `get_payment_brand` function handles this correctly, storing the `bank_name` in the `brand` field.

squarelet/organizations/tasks.py
```diff
@@ -430,13 +424,104 @@ def handle_customer_updated(customer_data):
...
+        )
+        return
+    # Retrieve the Stripe customer to verify this PM is the default.
+    # customer.updated will also fire in this scenario, so this is
+    # idempotent defense-in-depth.
+    customer_svc = get_payment_provider().get_customer_service()
+    stripe_customer = customer_svc.retrieve(customer_id)
+    invoice_settings = getattr(stripe_customer, "invoice_settings", None)
+    default_pm_id = invoice_settings and invoice_settings.get("default_payment_method")
+    if default_pm_id != stripe_id:
```
allanlasser: I'm not sure I follow the logic here. If the payment method is being attached, what's the likelihood it's the default payment method? Even though we can support multiple payment methods, we're throwing this away and creating inconsistency between Stripe and Accounts.

mitchelljkotler: Although the data model supports saving multiple payment methods, this isn't exposed or used anywhere.  We could save the data locally to keep things in sync, but don't want to increase the scope to include full multiple payment method handling support at this time.

mitchelljkotler: I added support to save non default payment methods for data syncing

allanlasser (approved):

PR #748, opened 2026-07-22
Stripe receipts

### LINKED ISSUE #91: Better itemize receipts - use Stripe Receipts
From @amandabee 
> I think it would be helpful to do a somewhat better job of itemizing our receipts so that you’d see something like:
> * Organization Plan $100
> * Additional Users 7 @ $10 each— $70
> * Prorated User [calculated]

I think we should consider using Stripe's built in receipts

mitchelljkotler:
This switches all receipts going forward to Stripe generated receipts.  We will continue to show pre-existing receipts as they are.  New receipts will be email by Stripe as well as stored by us indefinitely.

Stripe only supports a single email to send the receipt to, so we will no longer support multiple receipt emails.  We need to let anybody who relies on this feature know about this before hand.


allanlasser:
We don't need to do additional work on top of what we've done in #716 beyond fixing compatibility for the new receipt links and the change from multiple receipt emails to a single billing email.

dnass (approved):

eyeseast (approved):

PR #749, opened 2026-07-23
Add 'election' tier to Wix label aliases

allanlasser:
Wix has a character limit on labels, so we can't use the plan slug as our label. We need custom handling that turns `election-accountability-cohort` into `election`.


duckduckgrayduck:
Approving this PR to get us unstuck. Something that Mitch suggested in my handling of custom tags for certain affiliations (RNN/GSNC since I couldn't use the org name or slug) is to create a mapping and store it as an environment variable dict 
https://django-environ.readthedocs.io/en/latest/tips.html#complex-dict-format
We can open this as a separate issue. This would allow us to make these changes from the Heroku environment variables instead of requiring code changes

PR #750, opened 2026-07-23
un-cancel a subscription

### LINKED ISSUE #736: Give users a way to enable renew a canceled subscription
As mentioned in #732:

> If the subscription is active and cancelled, we can provide a way for the user to opt-in to renewal. This would unset the `cancelled` flag and update the `cancel_at_period_end` to `False`. We could encourage users to provide a payment method (CC or bank account) when setting up renewal. I think this makes sense as a standalone follow up to https://github.com/MuckRock/squarelet/issues/677.

If a subscription is canceled but we haven't yet reached the cancelation date, we should give users the ability to "uncancel" and renew their subscription. This will also need to unset the `cancel_at` date added in #735.

mitchelljkotler:
Addresses issue #736


allanlasser (changes requested):
We should check for an existing payment method on the account when resuming a subscription. If one doesn't exist, show an error message and redirect to the new `organizations/<slug>/card` route.

squarelet/organizations/payments/providers/stripe_modern.py
```diff
@@ -184,6 +184,13 @@ def cancel_at_period_end(self, stripe_subscription):
             cancel_at_period_end=True,
         )
 
+    def uncancel(self, stripe_subscription):
+        return stripe.Subscription.modify(
+            stripe_subscription.id,
+            cancel_at_period_end=False,
+            cancel_at="",
+        )
```
dnass, line 191: This throws an error: `Received both cancel_at_period_end and cancel_at parameters. Please pass in only one.` Removing either one seems to fix it, but I’m not sure if it makes a difference on Stripe's end.

dnass (approved):
Looks good!

allanlasser (approved):

PR #752, opened 2026-07-24
Prompt last admin to reassign before leaving

### LINKED ISSUE #564: Handle the last admin leaving an organization
[As discussed on 10/21/25](https://www.notion.so/Project-Planning-292f889269638130954dea8b7f1d7460?pvs=53):

- When the last admin leaves an organization, ask them if they want to reassign admin status or leave the org admin-less.
- If they choose to leave it admin-less, we create a ticket for staff: “The last admin left an organization—heads up.”

dnass:
Closes #564.

When the last admin leaves an organization or demotes themselves, they are shown a prompt with an opportunity to assign admin status to another member. If they decline, a Zendesk ticket is created.

image

Also simplifies organization `Detail.handle_leave` and its corresponding tests.

### DIFF
--- a/frontend/components/OrgUserSelect.svelte
+++ b/frontend/components/OrgUserSelect.svelte
@@ -1,0 +1,143 @@
+<script lang="ts">
+  import type { User } from "@/types";
+
+  import Svelecte from "svelecte";
+  import UserListItem from "./UserListItem.svelte";
+
+  interface Props {
+    organizationSlug: string;
+    organizationName: string;
+    onChange?: (selection: User) => void;
+  }
+
+  let { onChange, organizationName, organizationSlug }: Props = $props();
+
+  let value: User = $state();
+  const fetchProps: RequestInit = { credentials: "include" };
+
+  function handleChange() {
+    onChange?.(value);
+  }
+
+  /** Svelecte doesn't catch AbortError when it cancels in-flight fetches. */
+  function suppressAbortError(e: PromiseRejectionEvent) {
+    if (e.reason instanceof DOMException && e.reason.name === "AbortError") {
+      e.preventDefault();
+    }
+  }
+</script>
+
+<svelte:window onunhandledrejection={suppressAbortError} />
+
+<Svelecte
+  name="userid"
+  placeholder="Search for a member of {organizationName}..."
+  bind:value
+  valueAsObject
+  valueField="id"
+  labelField="name"
+  fetch="/fe_api/users/?organization={organizationSlug}"
+  {fetchProps}
+  fetchCallback={({ results }) => results}
+  fetchDebounceTime={400}
+  minQuery={0}
+  fetchResetOnBlur={false}
+  resetOnBlur={false}
+  lazyDropdown={false}
+  onChange={handleChange}
+  class="svelecte-control user-search"
+  --sv-min-height="2rem"
+  --sv-disabled-bg="var(--gray-1, #f5f6f7)"
+  --sv-border="1px solid var(--gray-3, #99a8b3)"
+  --sv-border-radius="0.5rem"
+  --sv-placeholder-color="var(--gray-3, #99a8b3)"
+  --sv-icon-color="var(--gray-3, #99a8b3)"
+  --sv-icon-color-hover="var(--gray-4, #5c717c)"
+  --sv-separator-bg="var(--gray-2, #d8dee2)"
+  --sv-dropdown-border="1px solid var(--gray-2, #d8dee2)"
+  --sv-dropdown-shadow="var(--shadow-2, 0 6px 8px 0px rgba(30 48 56 / 0.1))"
+  --sv-dropdown-active-bg="var(--blue-1, #eef3f9)"
+  --sv-dropdown-selected-bg="var(--blue-1, #eef3f9)"
+  --sv-loader-border="2px solid var(--blue-3, #4294f0)"
+>
+  {#snippet option(item: User)}
+    <UserListItem user={item} />
+  {/snippet}
+
+  {#snippet selection(selectedOptions: User[], bindItem)}
+    {#each selectedOptions as sel (sel.id)}
+      <div class="chip">
+        {sel.name || sel.username}
+        <button data-action="deselect" use:bindItem={sel}>&times;</button>
+      </div>
+    {/each}
+  {/snippet}
+</Svelecte>
+
+<style>
+  :global(.sv-control--selection) {
+    padding: 0 0.25em;
+  }
+
+  :global(.svelecte-control.user-search) {
+    font-family: var(--font-sans, "Source Sans Pro"), sans-serif;
+    font-size: var(--font-md, 1rem);
+    font-feature-settings: "ss04" on;
+  }
+
+  :global(.svelecte-control.user-search input) {
+    font-family: var(--font-sans, "Source Sans Pro"), sans-serif;
+    font-size: var(--font-md, 1rem);
+  }
+
+  :global(.svelecte-control.user-search .sv-dropdown) {
+    border-radius: 0.5rem;
+  }
+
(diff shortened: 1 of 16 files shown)


frontend/components/OrgUserSelect.svelte
```diff
@@ -0,0 +1,143 @@
...
+  function suppressAbortError(e: PromiseRejectionEvent) {
+    if (e.reason instanceof DOMException && e.reason.name === "AbortError") {
+      e.preventDefault();
+    }
+  }
+</script>
+
+<svelte:window onunhandledrejection={suppressAbortError} />
+
+<Svelecte
```
dnass: This should incorporate the new base Select component from #738 but it's not a blocker. Once one of these PRs is merged, I can update the other accordingly.

allanlasser: I just merged #738, so this should be ready for rebase+update!

dnass: All set

dnass:
The E2E tests seem to be failing at 10 minutes even though the [workflow timeout](https://github.com/MuckRock/squarelet/actions/runs/30276111733/workflow#L14) is set to 30.

squarelet/users/fe_api/viewsets.py
```diff
@@ -29,6 +29,9 @@ def get_queryset(self):
                 return User.objects.none()
             qs = User.objects.get_searchable(self.request.user).filter(is_active=True)
             search = self.request.query_params.get("search", "").strip()
+            org = self.request.query_params.get("organization", "").strip()
+            if org:
+                qs = qs.filter(organizations__slug=org)
```
allanlasser: This has the potential to leak the member list of an organization a user doesn't have the ability to view. We should first get the org queryset,  `Organization.objects.filter(slug=org).get_viewable(self.request.user)`. If that list is empty, we can return an empty user list. If not, we can filter the `User` queryset `qs` against that list.

squarelet/users/fe_api/test_user_viewsets.py
```diff
@@ -136,3 +136,101 @@ def test_search_list_fields(client, user_with_org):
...
+    outsider.individual_organization.hidden = False
+    outsider.individual_organization.save()
+    Membership.objects.create(user=outsider, organization=other_org)
+
+    client.force_authenticate(user=user)
+    response = client.get("/fe_api/users/")
+    assert response.status_code == status.HTTP_200_OK
+    returned_ids = {u["id"] for u in response.data["results"]}
+    assert member.id in returned_ids
+    assert outsider.id in returned_ids
```
allanlasser, line 209: Let's add a test to ensure that we're only returning users for orgs the user can view. Filtering the viewset by a private org the user doesn't belong to should return an empty list.

squarelet/organizations/views/members.py
```diff
@@ -395,3 +423,70 @@ class OrgRequestsView(BaseOrgInvitationRequestView):
...
+                messages.error(request, _("User is not a member of this organization"))
+                return redirect(f"organizations:{url}", org.slug)
+        else:
+            # If there's no replacement admin, create a Zendesk ticket
+            manage = reverse("organizations:manage-members", args=[org.slug])
+            manage_link = request.build_absolute_uri(manage)
+            subject = f"Organization with no admins: {org.name}"
+            description = (
+                f"The last admin has left {org.name}. ",
+                f"Follow this link to assign a new admin: {manage_link}",
```
allanlasser: I _think_ the presence of both commas turns this into a tuple object, instead of concatenating the string. I believe the commas need to be omitted here.

squarelet/organizations/views/members.py
```diff
@@ -395,3 +423,70 @@ class OrgRequestsView(BaseOrgInvitationRequestView):
...
+
+        membership = user.memberships.get(organization=org)
+
+        if is_leave_form:
+            membership.delete()
+            messages.info(request, _("You left the organization"))
+        else:
+            membership.admin = False
+            membership.save()
+            messages.info(request, _("You are demoted to a member"))
```
allanlasser:
```suggestion
            messages.info(request, _("You are no longer an admin"))
```

squarelet/organizations/views/members.py
```diff
@@ -395,3 +423,70 @@ class OrgRequestsView(BaseOrgInvitationRequestView):
...
+        url = "reassign-admin" if is_leave_form else "reassign-admin-demote"
+
+        if new_admin_id == str(user.id):
+            messages.warning(request, _("You must assign a user other than yourself"))
+            return redirect(f"organizations:{url}", org.slug)
+
+        if new_admin_id:
+            # If the outgoing admin picked a replacement, promote them
+            try:
+                membership = org.memberships.get(user_id=new_admin_id)
```
allanlasser: The repeated assignment of `membership` here and on line 478 is slightly confusing—can we name this `replacement_admin` or similar?

squarelet/organizations/views/members.py
allanlasser: These two transactions (new admin, old admin) should be wrapped in `with transaction.atomic():`, so that if an error happens during deletion/demotion, the operation aborts instead of leaving the other admin promoted.

frontend/components/OrgUserSelect.svelte
```diff
@@ -0,0 +1,56 @@
...
+<svelte:window onunhandledrejection={suppressAbortError} />
+
+<Select
+  name="userid"
+  placeholder="Search for a member of {organizationName}..."
+  bind:value
+  valueAsObject
+  valueField="id"
+  labelField="name"
+  fetch="/fe_api/users/?organization={organizationSlug}"
```
allanlasser, line 39: May be an edge case, but for an org with >100 members, this will only make the first page of members available as options. Passing through the query should help reveal members missing from the original list as the user types:

```suggestion
  fetch="/fe_api/users/?organization={organizationSlug}&search=[query]"
```

allanlasser (changes requested):

frontend/components/OrgUserSelect.svelte
dnass: Svelecte has an annoying limitation here: it doesn't perform an initial fetch when `[query]` is present in the fetch URL, so the list will be empty instead of complete when it's first opened. It also doesn't expose a binding or onchange handler for its internal input value, so we can't even work around it by interpolating the URL ourselves. 

Is the tradeoff worth it? How many orgs have >100 members and only one admin?

squarelet/users/fe_api/viewsets.py
dnass: Thanks for catching this. fixed and added a test

frontend/components/OrgUserSelect.svelte
allanlasser: Oh, that is so annoying! I don't think the tradeoff is worth it.

Maybe we can add a sort parameter into the viewset and use some kind of relevance signal—I would expect an org with >100 members to have many inactive users. What do you think about sorting the list by last login date, annotated by the `LoginLog` relationship?

By the way, we added the private `fe_api` viewsets API so we could modify the endpoint without worrying about impacting clients. Probably obvious, but worth calling out.

dnass: The [docs](https://svelecte.vercel.app/fetch#user-provided-fetch-function:~:text=User%2Dprovided%20fetch,every%20use%20case.) just rub salt in the wound.

I'll add the last login sorting now.

dnass: Looks like `LoginLog`s only track sign-ins on third-party clients. I think we can just sort on `user.last_login`?

dnass: Just confirmed that logging in on a client also updates the Squarelet `last_login` value, so it should always reflect the most recent first- or third-party login.

allanlasser (approved):

PR #754, opened 2026-07-27
Consolidate benefits display into single card

### LINKED ISSUE #753: Consolidate inherited subscriptions/benefits
Now that an organization can have multiple subscriptions, the "inherited benefits" section of user/org detail pages can grow unwieldy:

Instead of subscription-by-subscription, let's consolidate this into a single card that shows _all_ the benefits a user/affliate org can access through inheritance. This will change how we model `benefits` from a JSON blob on a plan to an array of strings on entitlements, providing a "plain english" translation of the entitlements blob.

allanlasser:
Closes #753 

- Adds new "benefits" field to `Entitlement`, which `Plan` can compose into a single list.
- Falls back to existing `Plan.benefits` value, deduplicating strings when possible.
- Shows a single "As a member of Org A, Org B, you inherit: <benefits list>"
- Shows the same consolidated benefits list for multiple owned plans.


squarelet/organizations/models/organization.py
```diff
@@ -967,3 +967,22 @@ def merge(self, org, user):
...
+    benefits card instead of one card per inherited plan.
+    """
+    orgs = []
+    benefits = []
+    for org, plan in inherited_plans:
+        if org not in orgs:
+            orgs.append(org)
+        for benefit in plan.get_benefits():
+            if benefit not in benefits:
+                benefits.append(benefit)
```
dnass: For benefits that involve quantities, it would be great to sum them. String deduping could introduce confusion in a couple ways: If a user has two plans that each provide 50 MuckRock requests, the consolidated list will show one entry that says 50 instead of 100. And if they have one plan that provides 10 requests and another that provides 50, they'll see two separate list entries and it may not be clear whether they're additive or one supersedes the other.

`benefits` is a `JSONField` so this could be accomplished without changing the model. Benefit entries could look like `{ text: "%(quantity)d free public records requests each month", quantity: 50 }` and then we'd need to introduce a couple new methods for summing and formatting.

squarelet/templates/organizations/includes/plan_card.html
```diff
@@ -93,44 +94,41 @@ <h3>{% trans "Your plans" %}</h3>
...
+        {% trans "You&rsquo;re inheriting these benefits as a member of" %}
       {% else %}
-        {% trans "Inherits" %}
+        {% trans "Inherits these benefits as a member of" %}
       {% endif %}
-      <a class="plan-name"
-         href="{% url 'plan_detail' pk=source_plan.id slug=source_plan.slug %}">{{ source_plan.name }}</a>
-      {% trans "benefits as a member of" %}
-      <a class="org-name" href="{{ source_org.get_absolute_url }}">{{ source_org.name }}</a>.
+      {% for org in inherited_orgs %}<a class="org-name" href="{{ org.get_absolute_url }}">{{ org.name }}</a>{% if not forloop.last %}, {% endif %}{% endfor %}:
```
dnass: [This template](https://github.com/MuckRock/squarelet/blob/a61f8979805d67dc3e853e1a7b52d3a82d032899/squarelet/templates/users/user_detail.html#L170) has a nice pattern for joining a list with commas and "and"

squarelet/organizations/models/organization.py
dnass: I'm not totally clear on where/how the actual quantities associated with the benefits are stored. Maybe there's a way to draw on that data as the source of truth?

allanlasser: That's a great point. The quantities are stored in the `Entitlements.resources` JSON blob—there's no strict schema for entitlements.

dnass: So we can aggregate the `resources` objects, and the benefits can stay as deduped plain strings with named format arguments. Then we'd just pass `resources` as the format argument to each benefit string before rendering. Like `"{base_requests} free public records requests each month".format(**resources)`

squarelet/templates/organizations/includes/plan_card.html
```diff
@@ -93,44 +94,41 @@ <h3>{% trans "Your plans" %}</h3>
...
+        {% trans "You&rsquo;re inheriting these benefits as a member of" %}
       {% else %}
-        {% trans "Inherits" %}
+        {% trans "Inherits these benefits as a member of" %}
       {% endif %}
-      <a class="plan-name"
-         href="{% url 'plan_detail' pk=source_plan.id slug=source_plan.slug %}">{{ source_plan.name }}</a>
-      {% trans "benefits as a member of" %}
-      <a class="org-name" href="{{ source_org.get_absolute_url }}">{{ source_org.name }}</a>.
+      {% for org in inherited_orgs %}<a class="org-name" href="{{ org.get_absolute_url }}">{{ org.name }}</a>{% if not forloop.last %}{% if forloop.revcounter == 2 %} and {% else %}, {% endif %}{% endif %}:
```
dnass:
```suggestion
      {% for org in inherited_orgs %}<a class="org-name" href="{{ org.get_absolute_url }}">{{ org.name }}</a>{% if not forloop.last %}{% if forloop.revcounter == 2 %} and {% else %}, {% endif %}{% endif %}{% endfor %}:
```

squarelet/organizations/views/detail.py
```diff
@@ -81,9 +86,11 @@ def get_context_data(self, **kwargs):
             org.subscriptions.filter(plan__wix=True).exists()
             or org.get_wix_plans_from_groups()
         )
-        context["inherited_plans"] = org.get_inherited_plans()
-
-        context.update(self._get_groups_context(user, org))
+        inherited_orgs, inherited_benefits = consolidate_inherited_benefits(
+            org.get_inherited_plans()
```
dnass, line 101: We're ONLY showing inherited benefits here. If an org has its own subscription, the benefits don't appear anywhere on the plan card. Seems like those should get rolled into the consolidated benefits too.

squarelet/users/views.py
```diff
@@ -176,11 +181,15 @@ def get_context_data(self, **kwargs):
...
-        # paid plan. Plans the org inherits from parents/groups are intentionally
-        # excluded here -- those are an org-level concept and confuse users when
-        # surfaced on their personal page.
-        context["premium_org_plans"] = self._get_premium_org_plans(user)
+        # Consolidated benefits from each non-individual org the user belongs to
+        # that has its own paid plan. Plans the org inherits from parents/groups
+        # are intentionally excluded here -- those are an org-level concept and
+        # confuse users when surfaced on their personal page.
+        inherited_orgs, inherited_benefits = consolidate_inherited_benefits(
+            self._get_premium_org_plans(user)
```
dnass, line 198: Same here — if a user has an individual subscription, the benefits aren't included on the plan card.

squarelet/users/views.py
dnass: This doesn't seem to handle member orgs that inherit plans from a parent, only orgs with their own direct subscriptions.

squarelet/organizations/views/detail.py
allanlasser: Good catch! I'd prefer to benefits _you/this org paying for_ separated from inherited benefits, but we can totally summarize and consolidate multiple owned subscriptions in the same way.

squarelet/users/views.py
dnass: There's still an issue here with multiple levels of benefit inheritance. Here's an org that has its own subscription plus an inherited subscription via an affiliation:

Image

A user in this org only sees the inherited benefits from the org's direct subscription, not the org's own inherited benefits:
Image

squarelet/templates/organizations/includes/plan_card.html
```diff
@@ -93,44 +117,41 @@ <h3>{% trans "Your plans" %}</h3>
...
-{% for source_org, source_plan in inherited_plans %}
+{% if inherited_benefits %}
   <main class="plan-card plan-card--inherited">
     <p class="plan-overview">
       {% if subject == "users" %}
-        {% trans "You inherit" %}
+        {% trans "You&rsquo;re inheriting these benefits as a member of" %}
       {% else %}
-        {% trans "Inherits" %}
+        {% trans "Inherits these benefits as a member of" %}
```
dnass: Should this be "as an affiliate of"?

squarelet/templates/organizations/includes/plan_card.html
```diff
@@ -71,6 +74,27 @@ <h3>{% trans "Your plans" %}</h3>
         {% endfor %}
       </ul>
     </div>
+
+    {% if subscription_benefits %}
+      <div class="plan-detail">
```
dnass: Feels like either the row gaps should be uniform here or `Included benefits` should be visually offset some other way, like in a nested box.

Image

squarelet/users/views.py
dnass: There's a comment that says:

> Plans the org inherits from parents/groups are intentionally excluded here -- those are an org-level concept and confuse users when surfaced on their personal page

but that's from before we were consolidating the benefits. IMO it's now more confusing NOT to include inherited benefits. In this context, users have no reason to know or care whether they have 50 requests because of their org's subscription or their org's affiliate's subscription — they should see the full total of benefits available to them.

allanlasser:
While updating the design of the billing card, I did away with many of the action links, which all pointed to the exact same place. I found this keeps the info easier to parse and more summary-like.

Billing card for a user with multiple subscriptions who is also a member of an org with multiple subscriptions:

Screenshot 2026-09-03 at 19 07 57

Billing card for an organization with multiple subscriptions and inherited benefits from an affiliation:

Screenshot 2026-09-03 at 19 07 52

dnass (approved):
The new plan card layout and language are super clear. Looks great.

PR #756, opened 2026-07-28
Look up change requests to redirect on old org slugs

### LINKED ISSUE #562: Handle updates to organization slugs
Like with individual profiles, slugs may be changed once before editing them is disabled. If an org admin would like to change their slug again, they need to contact support with a change request. With #506, it's now easier for organizations to request changes to their URL slugs.

We should provide handling for when org slugs change:

1. When we can’t find an org matching a slug, we can query the table of approved change requests to find an older slug that matches. If we find a match, we can respond with a 301 redirect to the organization's current slug.
2. Since changing a slug will break existing URLs to the organization’s pages in Accounts, we can handle broken links by showing signed-in users a list of their orgs on any organization 404 pages.

dnass:
Closes #562.

Introduces `ResolveOrganizationSlugMixin`, which tries to match the provided org slug against an accepted profile change request and returns a 301 request if one is found. Currently this is only used with the organization `Detail` view, but it could be added to other org views too.

Also adds a custom 404 handler, and shows signed-in users a list of their orgs on organization 404 pages.

To test the 404 page in context (i.e. `/organizations/not-a-real-org/` rather than `/404`) I had to tweak `config/settings/local.py` locally:

```diff
-DEBUG = True
+DEBUG = False

DJANGO_VITE = {
    "default": {
-       "dev_mode": DEBUG,
+       "dev_mode": True,
```

### DIFF
--- a/squarelet/organizations/mixins.py
+++ b/squarelet/organizations/mixins.py
@@ -6,6 +6,7 @@ from django.shortcuts import redirect
 # Squarelet
 from squarelet.core.exceptions import ContextHttp404
 from squarelet.organizations.models import ProfileChangeRequest
+from squarelet.organizations.models.organization import Organization
 
 
 class OrganizationAdminMixin(UserPassesTestMixin):
@@ -83,16 +84,19 @@ class ResolveOrganizationSlugMixin:
         slug = self.kwargs.get("slug")
         change_request = (
             ProfileChangeRequest.objects.filter(previous__slug=slug, status="accepted")
+            .exclude(organization__slug=slug)
             .select_related("organization")
             .order_by("-updated_at")
             .first()
         )
         if change_request is None:
             return None
-        org = change_request.organization
-        if not user.has_perm("organizations.view_organization", org):
+        try:
+            pk = change_request.organization.pk
+            org = Organization.objects.get_viewable(user).get(pk=pk)
+            return redirect(org.get_absolute_url(), permanent=False)
+        except Organization.DoesNotExist:
             return None
-        return redirect(org.get_absolute_url(), permanent=True)
 
     def get_user_orgs(self, request):
         user = request.user
--- a/squarelet/organizations/mixins.py
+++ b/squarelet/organizations/mixins.py
@@ -1,5 +1,10 @@
 # Django
 from django.contrib.auth.mixins import PermissionRequiredMixin, UserPassesTestMixin
+from django.http import Http404
+from django.shortcuts import redirect
+
+# Squarelet
+from squarelet.organizations.models import ProfileChangeRequest
 
 
 class OrganizationAdminMixin(UserPassesTestMixin):
@@ -55,3 +60,32 @@ class VerifiedJournalistMixin(UserPassesTestMixin):
             and organization.verified_journalist
             and organization.has_admin(self.request.user)
         )
+
+
+class ResolveOrganizationSlugMixin:
+    """Resolve an org by slug, or look for a matching change request
+    and return a redirect if one is found."""
+
+    def get(self, request, *args, **kwargs):
+        try:
+            self.object = self.get_object()
+            context = self.get_context_data(object=self.object)
+            return self.render_to_response(context)
+        except Http404:
+            slug_redirect = self.get_slug_redirect()
+            if slug_redirect is not None:
+                return slug_redirect
+            raise
+
+    def get_slug_redirect(self):
+        slug = self.kwargs.get("slug")
+        change_request = (
+            ProfileChangeRequest.objects.filter(previous__slug=slug, status="accepted")
+            .select_related("organization")
+            .order_by("-updated_at")
+            .first()
+        )
+        if change_request is None:
+            return None
+        org = change_request.organization
+        return redirect(org.get_absolute_url(), permanent=True)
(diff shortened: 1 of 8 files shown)


squarelet/organizations/mixins.py
```diff
@@ -55,3 +61,40 @@ def test_func(self):
...
+        except Http404:
+            slug_redirect = self.get_slug_redirect(request.user)
+            if slug_redirect is not None:
+                return slug_redirect
+            raise ContextHttp404(context={"user_orgs": self.get_user_orgs(request)})
+
+    def get_slug_redirect(self, user):
+        slug = self.kwargs.get("slug")
+        change_request = (
+            ProfileChangeRequest.objects.filter(previous__slug=slug, status="accepted")
```
allanlasser, line 86: When we create new ProfileChangeRequests, we copy all existing org values into `previous`, not just the values that are requesting changed. An organization that changes other values, but not slug, will still return a record here, even though the slug is the same. This will enter into a 301 redirect loop when we redirect, always redirecting to the same slug.

Since we trigger the redirect on 404, we'll enter this loop when a user without permission to view an org that's changed some profile data, but not it's slug — private orgs being the most obvious case.

I think the right solution is adding `.exclude(organization__slug=slug)` to break the loop.

squarelet/organizations/mixins.py
```diff
@@ -55,3 +61,40 @@ def test_func(self):
...
+            .select_related("organization")
+            .order_by("-updated_at")
+            .first()
+        )
+        if change_request is None:
+            return None
+        org = change_request.organization
+        if not user.has_perm("organizations.view_organization", org):
```
allanlasser: The permission for `view_organization` is `is_public | is_member`, but we still want staff members to have visiblity into orgs, too. Staff would get at 404 instead of a redirect here.

I believe this will also leak the new slug for an organization that we're intentionally hiding for moderation purposes—we intentionally hide _public_ organizations that aren't verified and don't have any payment history with us. If a hidden org has a name/slug change on record, we'll redirect users to an org we don't want them to be able to see.

squarelet/organizations/mixins.py
```diff
@@ -55,3 +61,40 @@ def test_func(self):
...
+            return None
+        return redirect(org.get_absolute_url(), permanent=True)
```
allanlasser: With a 301 redirect, it's hard for another organization to "claim" a slug released by another. I think a temporary redirect is probably the better option here, so it's not cached long-term.

squarelet/organizations/mixins.py
```diff
@@ -55,3 +61,40 @@ def test_func(self):
...
+
+    def get_user_orgs(self, request):
+        user = request.user
+        if not hasattr(user, "organizations"):
+            return []
+        return user.organizations.filter(individual=False)
```
allanlasser: Let's apply some kind of sort to this list—alphabetical is probably best.

squarelet/organizations/mixins.py
```diff
@@ -55,3 +61,40 @@ def test_func(self):
...
+class ResolveOrganizationSlugMixin:
+    """Resolve an org by slug, or look for a matching change request
+    and return a redirect if one is found."""
+
+    def get(self, request, *args, **kwargs):
+        try:
+            self.object = self.get_object()
+            context = self.get_context_data(object=self.object)
+            return self.render_to_response(context)
+        except Http404:
```
allanlasser: We should move the two calls outside this `try`, so that we don't mishandle a context or rendering error as a "org can't be found by slug" error.

squarelet/templates/404.html
```diff
@@ -1,11 +1,37 @@
...
+      {% if user_orgs|length %}
+        <h2>{% trans "Were you looking for one of these organizations?" %}</h2>
+        <ul>
+          {% for org in user_orgs %}
+            <li>
+              <a class="btn ghost primary" href="{{ org.get_absolute_url }}">{{ org.name }}</a>
+            </li>
+          {% endfor %}
+        </ul>
+      {% endif %}
```
allanlasser, line 30: Could use an empty state for the helpful wayfinding here. We could render a link back to `/organizations` for a user ("Go to organizations overview"), displaying it in just the `else` case or in both cases underneath the conditional.

squarelet/organizations/tests/test_views.py
```diff
@@ -526,6 +539,113 @@ def test_post_staff_remove_nonexistent_user(
         # Should still redirect but show an error message
         assert response.status_code == 302
 
+    def test_redirects_old_slug_after_accepted_rename(
+        self, rf, user_factory, organization_factory, profile_change_request_factory
+    ):
+        user = user_factory()
+        organization = organization_factory(name="Old Newsroom", users=[user])
```
allanlasser, line 546: `organization_factory` generates verified newsroom orgs by default—it'd be good to ensure our handling for other org types, especially the "quietly hidden" orgs that are public but neither verified nor have a payment history.

allanlasser (changes requested):
Overall this looks great. Caught some issues in the mixin logic and left some suggestions on template and tests.

There's some complex logic around org visibility (related to moderation and spam reduction) that's not well documented, so it makes sense that it got overlooked in a first pass.

allanlasser (approved):

PR #759, opened 2026-08-05
Improve observability of cache invalidation webhooks

allanlasser:
- Adds a custom logger for `squarelet.oidc` to capture INFO in addition to WARNING and ERROR
- Raises errors when webhook tasks return 4xx or 5xx responses.
- Logs when client consent filters lead to no-op cache invalidation

### DIFF
--- a/squarelet/oidc/models.py
+++ b/squarelet/oidc/models.py
@@ -7,10 +7,38 @@ from django.utils.translation import gettext_lazy as _
 # Standard Library
 import hashlib
 import hmac
+import logging
 import time
 
 # Third Party
 import requests
+
+logger = logging.getLogger(__name__)
+
+# (connect, read) - a stuck TLS handshake should not pin a worker for the full
+# read timeout
+WEBHOOK_TIMEOUT = (5, 30)
+# clients return full debug HTML on error, and a sustained outage logs this on
+# every attempt of every broadcast - enough to identify the error, no more
+ERROR_BODY_CHARS = 200
+# Log every uuid for interactive saves, where the list itself is what you need to
+# trace a delivery. Bulk jobs like restore_organization pass thousands at once.
+UUID_LOG_THRESHOLD = 30
+UUID_LOG_SAMPLE = 3
+
+
+def get_elapsed_ms(start):
+    """Return elapsed ms since start"""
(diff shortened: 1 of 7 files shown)


squarelet/oidc/models.py
```diff
@@ -74,4 +106,49 @@ def send_cache_invalidation(self, model, uuids):
...
+            )
+            raise
+
+        elapsed_ms = get_elapsed_ms(start)
+        if not response.ok:
+            # log at warning, not error: the task logs a single error once it has
+            # exhausted its retries, so one failed delivery is one Sentry event
+            logger.warning(
+                "[CACHE-INVALIDATION] Rejected client=%s model=%s uuids=%s "
+                "status=%d elapsed_ms=%d body=%s",
```
mitchelljkotler: Do we want the URL in here too?

squarelet/oidc/models.py
```diff
@@ -74,4 +106,49 @@ def send_cache_invalidation(self, model, uuids):
...
+                model,
+                format_uuids(uuids),
+                response.status_code,
+                elapsed_ms,
+                response.text[:ERROR_BODY_CHARS],
+            )
+            response.raise_for_status()
+
+        logger.info(
+            "[CACHE-INVALIDATION] Sent client=%s model=%s count=%d status=%d "
```
mitchelljkotler: URL here too

squarelet/oidc/tasks.py
```diff
@@ -40,3 +48,14 @@ def send_cache_invalidation(client_profile_pk, model, uuids):
 
     if uuids:
         client_profile.send_cache_invalidation(model, uuids)
```
mitchelljkotler: This method now throws an error on any non-200 responses - these should either not be raised (they are logged), caught, or retried on.  Retrying on certain status codes might make sense.

mitchelljkotler (changes requested):
This looks good overall, some minor changes.  If you want to merge this to debug now, retry logic can be added on later if needed.

squarelet/oidc/tasks.py
mitchelljkotler: Also, do you want to pass `invalidation_id` in here for better traceability?

mitchelljkotler (approved):

PR #760, opened 2026-08-05
Audit subscriptions

mitchelljkotler:
This is a command which compares active subscriptions with the data on Stripe and reports any discrepancies.  Useful as we try to consolidate our plans and get everything into sync.

### DIFF
--- a/squarelet/organizations/management/commands/audit_subscriptions.py
+++ b/squarelet/organizations/management/commands/audit_subscriptions.py
@@ -12,13 +12,14 @@ from datetime import datetime, timezone as dt_timezone
 import stripe
 
 # Squarelet
-from squarelet.organizations.models.payment import Subscription
+from squarelet.organizations.models.payment import Customer, Subscription
 
 logger = logging.getLogger(__name__)
 
 stripe.api_key = settings.STRIPE_SECRET_KEY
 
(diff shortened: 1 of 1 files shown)


squarelet/organizations/management/commands/audit_subscriptions.py
```diff
@@ -0,0 +1,303 @@
...
+            )
+
+        # cancel_at (date)
+        cancel_at_ts = getattr(stripe_sub, "cancel_at", None)
+        stripe_cancel_at = (
+            datetime.fromtimestamp(cancel_at_ts, tz=dt_timezone.utc).date()
+            if cancel_at_ts
+            else None
+        )
+        if sub.cancel_at != stripe_cancel_at:
```
duckduckgrayduck, line 263: sub.cancel_at is stored in EDT/EST. I get false alarms when I run this script in the local Django shell because we are casting the Stripe sub's cancel at to UTC, but the sub.cancel_at is EDT/EST and so we get things like this

cancel_at: local=2027-06-09  stripe=2027-06-10

duckduckgrayduck:
Otherwise looks good. I do see a lot of matches for things like legacy stripe plan names. 

plan_stripe_id: local=squarelet_plan_professional  stripe=pro

duckduckgrayduck (approved):

Issue #654, opened 2026-04-20
Org search needs optimization


duckduckgrayduck:
The new org search (which relies on /fe_api/organizations) is reaching a lot of request timeouts. 
It is our #1 contributor to H12 codes on Heroku. In the logs I see searches like

?individual=false&search=Politic
?individual=false&search=Ohio%20University
?individual=false&search=Arkansa&fuzzy=true
timing out.

PR #761, opened 2026-08-11
Fix inefficient SQL queries in fe_api/organizations 

duckduckgrayduck:
Related Sentry error:
https://muckrock.sentry.io/issues/7273932599/?project=1422309&query=is%3Aunresolved&referrer=issue-stream

Started 6 months ago, which I now believe is related to [my own commit](https://github.com/MuckRock/squarelet/commit/f0ba70407cb42df6b600e6a18e928dbace05a1d6) to add back potential orgs a user can join back into the get_viewable in a rush after we released organization search. This is a big contributor to #654 which has since been closed. 

**What this did before:** 
qs is all orgs. For staff, we just return that and don't do any further stuff to the queryset. So when we test something on a staff account we don't notice any performance issues in org search. For non-staff, it ran the can_auto_join check along the entire queryset which checks if the org has auto join enabled and if the org is in the list of potential orgs: https://github.com/MuckRock/squarelet/blob/23470837fdc8f9219b166bd2a5a336610e9b8e21/squarelet/users/models.py#L287
This doesn't scale with the size of the queryset. can_auto_join is a permissions level guard when you're looking at a singular org to determine whether that user should be able to join it automatically. 

**What this does now:**
get_potential_organizations already does the necessary filter to give us the set of orgs that allow_auto_join and are available to the user. All we need to do is pass those orgs to the viewable filter. This passes the orgs as a sub-query of primary keys to the viewable filter so we don't have to hold all the orgs in memory. This should also be a lot more readable. 

I verified this in the shell against a non-staff user with a confirmed email address and some test orgs with auto join enabled and some domains set. The endpoint's number of SQL queries drop from 10 queries to 2.

### DIFF
--- a/squarelet/organizations/querysets.py
+++ b/squarelet/organizations/querysets.py
@@ -31,19 +31,14 @@ class OrganizationQuerySet(models.QuerySet):
             # or they can auto join that org
             # and they can only see public organizations that are visible
             # (verified or have charges or paid invoices)
+            auto_join_org_pks = user.get_potential_organizations().values("pk")
             viewable_filter = (
                 Q(private=False, verified_journalist=True)
                 | Q(private=False, charges__isnull=False)
                 | Q(private=False, invoices__status="paid")
                 | Q(users=user)
+                | Q(pk__in=auto_join_org_pks)
             )
-
-            # Include auto-join orgs (pre-approved) in the same filter
-            if hasattr(user, "can_auto_join"):
-                auto_join_pks = [org.pk for org in qs if user.can_auto_join(org)]
-                if auto_join_pks:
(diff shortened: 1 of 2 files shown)


allanlasser (approved):
It'd be quite nice if we could assert query efficiency in tests, but no idea how that'd work.

duckduckgrayduck:
> It'd be quite nice if we could assert query efficiency in tests, but no idea how that'd work.

Good point. I've added a test for get_viewable that acts a regression guard. It asserts that the number of queries when that method is called for a user is 2. If something down the line changes that and it grows a lot, we know there is something wrong.

PR #763, opened 2026-08-18
Audit subscriptions

mitchelljkotler:
Update the stripe audit and add a command to sync current period end, which was not being synced correctly.


duckduckgrayduck (approved):

duckduckgrayduck (changes requested):
Wait, actually, it looks like _sync_one with dry_run=True doesn't work

duckduckgrayduck:
Disregard last comment, was caching, this works. I was able to sync one of the subscriptions dates and run the command to audit the existing subs and it brought up a lot of mismatch, which I see why the other PR is necessary.

duckduckgrayduck (approved):

PR #771, opened 2026-08-24
update django-allauth

mitchelljkotler:


duckduckgrayduck:
Just needs check to pass and then it's good to merge

duckduckgrayduck (approved):

PR #775, opened 2026-08-25
update webhook handler for new api version

mitchelljkotler:
I realized that the webhook handler is still set up for the old api version, and sending data in that format.  We want to upgrade the webhook handlers to the same API version we use everywhere else.

### DIFF
--- a/squarelet/organizations/tasks.py
+++ b/squarelet/organizations/tasks.py
@@ -206,7 +206,8 @@ def handle_invoice_failed(invoice_data):
         )
     except Organization.DoesNotExist:
         try:
-            # plan was removed in 2025-03-31.basil; pricing.price_details is the new path
+            # plan was removed in 2025-03-31.basil;
+            # pricing.price_details is the new path
             line = invoice_data["lines"]["data"][0]
             plan_id = (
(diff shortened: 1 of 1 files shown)


squarelet/organizations/tasks.py
```diff
@@ -205,7 +205,17 @@ def handle_invoice_failed(invoice_data):
             customers__customer_id=invoice_data["customer"]
         )
     except Organization.DoesNotExist:
-        if invoice_data["lines"]["data"][0]["plan"]["id"] == "donate":
+        try:
+            # plan was removed in 2025-03-31.basil; pricing.price_details is the new path
+            line = invoice_data["lines"]["data"][0]
+            plan_id = (
+                line.get("pricing", {}).get("price_details", {}).get("price")
```
duckduckgrayduck: It looks like we should make sure that pricing.type is price_details according to the migration docs, as they may support different types in the future:
https://docs.stripe.com/changelog/basil/2025-03-31/invoice-pricing-configurations#impact

Also, pricing is a nullable object so we need to handle that or could receive an AttributeError
https://docs.stripe.com/api/invoice-line-item/object?api-version=2025-03-31.basil#invoice_line_item_object-pricing

PR #781, opened 2026-08-26
Pricing models

mitchelljkotler:
This creates a new PlanPrice model to re-organize our plans using Stripe's price feature.  Instead of offering multiple plans with the same functionality but different payment options, there is only one plan per functional plan with different PlanPrice's.  This covers monthly vs annual charging, comped accounts, non-profit pricing, and special negotiated rates.


mitchelljkotler:
Claude notes:

 Additive schema plus Stripe plumbing; changes nobody's bill. PlanPrice, Plan.product/stripe_product_id, the ensure_stripe_* methods, the admin, and consolidate_stripe_products.

  Highest-value checks: add a price through the admin and confirm it appears in Stripe with squarelet_variant metadata; try editing that price's amount and confirm it's refused; run consolidate_stripe_products twice and confirm the second run creates nothing.

---

I ran consolidate_stripe_products twice and it worked as expected.  I also ensured that existing price's have read only fields for fields which shouldn't change.

squarelet/organizations/admin.py
```diff
@@ -528,6 +533,82 @@ def get_subscription_renews(self, obj):
...
+            if name in self.fields:
+                self.fields[name].disabled = True
+                self.fields[name].widget = ReadOnlyValueWidget()
+                self.fields[name].help_text = (
+                    "Fixed once the price exists.  Supersede this row to "
+                    "charge something else."
+                )
+
+
+class PlanPriceInline(admin.TabularInline):
```
allanlasser: It'd be very helpful to include a link to the record in Stripe, if it exists--Invoice does something like this with a "View in Stripe" link. Can omit or disable the link when a comp row has no matching Stripe object.

squarelet/organizations/tests/factories.py
```diff
@@ -138,6 +138,20 @@ class ProfessionalPlanFactory(PlanFactory):
     for_groups = False
 
 
+class PlanPriceFactory(factory.django.DjangoModelFactory):
+    """A price under a plan.  Defaults to a paid monthly list price."""
+
+    plan = factory.SubFactory("squarelet.organizations.tests.factories.PlanFactory")
+    interval = "monthly"
+    label = "standard"
+    code = ""
```
allanlasser: Is `code` how we're representing discounts, or does this mean something else? Without documentation, it's hard to say this for sure.

squarelet/organizations/payments/providers/stripe_modern.py
```diff
@@ -316,6 +316,54 @@ def retrieve_product(self, product_id):
...
+        """
+        metadata = getattr(obj, "metadata", None)
+        if metadata is None:
+            return {}
+        to_dict = getattr(metadata, "to_dict", None)
+        return to_dict() if to_dict is not None else dict(metadata)
+
+    def find_product(self, slug):
+        for product in stripe.Product.list(active=True, limit=100):
+            if self._metadata(product).get("squarelet_plan_slug") == slug:
```
allanlasser, line 356: We are missing a `get_product` and `get_price` method here. Since we store a Stripe ID for products and prices, we can skip the list + loop and just return the object we're looking for.

squarelet/organizations/models/payment.py
```diff
@@ -1009,6 +1120,293 @@ def delete_stripe_plan(self):
...
+
+    `interval` and `label` are orthogonal: a nonprofit on an annual plan has
+    interval="annual" and label="nonprofit".
+
+    An individually negotiated rate is a price of its own, identified by
+    `code` rather than by a label.  Stripe has no negative coupon, so a rate
+    *above* list cannot be expressed as a discount on one - and a rate below
+    list is the same kind of thing, so both are handled the same way.
+    Coupons are kept for time-limited promotions, where expiry and redemption
+    limits are what is wanted.
```
allanlasser, line 1137: I can understand why this was the approach taken for creating arbitrary, one-off prices, especially for our enterprise plans. That makes me think we may want a relation between a price and an organization, like how we have for Plans, so we can make certain prices only available to certain orgs.

We've talked a lot about providing discounts through [Coupons and Promo Codes](https://docs.stripe.com/billing/subscriptions/coupons), which really should be managed through Stripe (I'm not convinced we need to store local copies of those, I'm thinking it over). I want to make sure that any [Discounts](https://docs.stripe.com/api/discounts/object) created by applying coupons/codes to subscriptions aren't forgotten. Will that be addressed by another PR in the stack?

We've talked about leaning on Stripe as the UI for our staff to manage payments, then having our system respond in kind. I understand that prices are immutable, so sync is _mostly_ a non-issue here. I don't see how we're responding to cases where new prices are created in the Stripe console. I think we need a webhook event handler for this.

eyeseast (approved):

squarelet/organizations/admin.py
mitchelljkotler: Agreed - I added links for all models which have a corresponding record in Stripe

squarelet/organizations/payments/providers/stripe_modern.py
mitchelljkotler: `find_product` is only used when there is no Stripe ID, which is used to ensure creating them on Stripe is safe and idempotent - if it crashed halfway through, it can find the corresponding product.  There is `retrieve_product` - added `retrieve_price` for consistency.

squarelet/organizations/models/payment.py
mitchelljkotler: Documentation here: https://app.notion.com/p/muckrock/How-new-Billing-Works-3d7f8892696381cb89a9dccb7d44cebc

Yes, we will need to add a relation between orgs and prices.

Coupons are used for temporary discounts.  If the coupon is managed via Stripe I am not sure what else would need to change - once the coupon is applied it should persist.

Prices locally require more fields than what Stripe stores - they are stored as metadata on the Stripe side.  We could require staff to fill out the metadata, but I think creating them locally in the Django Admin with proper fields would be easier.  Can keep this an open issue, but I do not think it blocks this release.

duckduckgrayduck:
When a PlanPrice is deleted, should we archive it on Stripe?

mitchelljkotler:
> When a PlanPrice is deleted, should we archive it on Stripe?

Yes, Price's can't be deleted, so they should be archived instead when they are superseded.  This functionality is in #807

mitchelljkotler:
I am planning on releasing this on Monday

PR #782, opened 2026-08-26
backfill subscriptions onto plan prices

mitchelljkotler:
This adds a command to backfill subscriptions into the new data model

Claude notes:

One command, no schema, no Stripe calls. The checks worth doing are the refusals — no --actor, unknown --actor, and running it before consolidate_stripe_products — because the command's whole safety story is that it stops rather than guessing.

Then spot-check the distinction it exists to make: a paying subscriber lands on the standard price, a comped organization lands on the comped price with granted_by filled in.


mitchelljkotler:
Do not review yet, this needs to be re-ordered

PR #784, opened 2026-08-27
Cancel one plan of several, not the whole subscription

mitchelljkotler:
Stacks on #811, which carries the split itself. **This is the behaviour change alone** — 15 files, one feature.

## What changed

Removing one plan from a multi-plan subscription dropped it from Stripe immediately, while removing an organization's only plan let them keep it until the period they had paid for ran out. Same button, different outcome depending on how many other plans the customer happened to hold.

`SubscriptionItem` now carries its own `cancelled`/`cancel_at` (`0085`), which the migration backfills from the parent so anyone mid-cancellation at deploy time is not told their plan renews. Cancelling a line flags it and leaves it billing — it still grants access until the period ends — and `uncancel` reverses it. The last *active* line still cancels the whole subscription, because Stripe has no per-item `cancel_at_period_end` and will not remove a subscription's only line. `restore_organization` enforces the rest, dropping each line from Stripe once its date arrives, with proration suppressed since it has already been paid for.

## The pair is written in one place

`Cancellable` owns `cancelled` and `cancel_at` for both models, and `push_cancellation_to_items` copies from the subscription rather than restating the values, so a parent and its lines cannot be written to disagree. Every bug this area has had came from those two fields being set half at a time — a flag with no date reads to the nightly sweep as "delete this tonight".

## Three subtleties worth reading for

- **The Stripe webhook pushes only when the subscription's own cancellation changed.** It fires on renewals and payment-method changes too, and Stripe knows nothing about a line the customer cancelled — so answering "is this subscription ending?" onto every line revived per-line cancellations on the next unrelated update.
- **A line joining a subscription that is already ending inherits the ending**, rather than advertising a renewal that is not coming and offering to cancel something about to vanish.
- **The purchase form and `add_subscription` now agree.** A cancelled line still occupies its plan — `unique_together` is `(subscription, plan)` — so reviving one is Resubscribe's job rather than a second line. The form used to offer organizations it could not serve.

## QA

- [x] **Cancel one line of a two-line subscription.** That line shows as ending, the other keeps renewing, and the Stripe subscription survives with both items until the date arrives.
- [x] **Cancel the last active line.** The whole subscription cancels at period end, and that line still reads as the customer's own cancellation rather than the subscription's — the subscription is ending *because of* it.
- [x] **Cancel every plan on a subscription, then press Resubscribe on one of them.** That plan comes back, the subscription renews again because there is an active plan on it, and the others stay cancelled. Press Resubscribe on a different one and you get that one instead. (Before this PR's last commit you got back whichever plan happened to be cancelled last, and the one you clicked stayed off.)
- [ ] **Let the sweep run on a line whose date has arrived.** It disappears from the Stripe subscription; the others keep billing.
- [x] **Add a plan to a subscription that is pending cancellation.** The subscription starts renewing again, locally *and* on Stripe, and the new line is not ending. The plan they cancelled still ends, on its original date — and **resubscribing afterwards does not bring it back**, because it is now ending in its own right rather than because its parent was. Adding a *free* plan, or a one-off, leaves the cancellation standing: neither is billed, so neither is a reason to renew.
- [x] **Cancel a line, then trigger any unrelated subscription update** (a payment-method change will do). The cancellation survives.
- [x] **Schedule a cancellation from the Stripe dashboard**, then lift it there. Every line reports the cancellation; lifting it brings back the lines that were still renewing when it landed, and leaves alone any the customer had cancelled themselves. This is the one thing `cancelled_by_subscription` decides.
- [ ] **Before deploying, cancel something; after deploying, open its billing page.** It still says "Ends on", with no offer to cancel again — `0085` backfills the line-level flags from the parent, and this is the only window in which that is observable.
- [ ] **A subscription mixing a free and a paid plan** bills correctly and does not error.
- [ ] **Cancel the paid plan on that mixed subscription** — from the site, and again from the Stripe dashboard. The free plan is not marked as ending either time, and when the date arrives the paid plan is gone and the free one is still there. A free plan was never sent to Stripe, so nothing that ends the Stripe subscription is about it; it used to be flagged along with everything else and then deleted with the subscription row.


mitchelljkotler:
Claude notes:

 The riskiest, and the checklist is longest. The core thing to prove is the one-invoice behaviour, in Stripe's dashboard rather than in squarelet:

  - Subscribe to a monthly plan → one Stripe subscription
  - Subscribe to a second monthly plan → must appear as a second line item on the same subscription, with one upcoming invoice
  - Subscribe to an annual plan → correctly a separate subscription, different billing shape

  Then the cancellation matrix: cancelling one line of two leaves the other billing and leaves both items on Stripe until period end; cancelling the only line sets cancel_at_period_end; resubscribe reverses each.

  And one interface check I'd not skip: diff an organization's OIDC entitlements payload before and after the migration. That's the contract MuckRock and DocumentCloud consume. It should be byte-identical — same entries, same resources, same quantity — but it's the thing that breaks other teams if I'm wrong.

  Two caveats on the checklists: they assume Stripe test keys in the preview environment, and #782/#784 both assume consolidate_stripe_products has been run there first, since the backfill needs the prices to exist.

mitchelljkotler:
Superseded by the release 3 redesign: grouped invoices with whole-subscription cancellation, one-offs on their own subscription, free and paid never mixed. The per-line cancellation code is archived at tag `archive/per-line-cancel/item-split`. The branch is left in place because later PRs still build on it until they're rebased.

PR #798, opened 2026-09-03
Step 2c — sell subscriptions against PlanPrice

mitchelljkotler:
Stacks on #784.

**Switches the sign-up flow onto `PlanPrice` *before* the migration, not after.** That's what lets #800 be a single run rather than a sweep that has to be repeated — every subscription taken out after this ships already has a price recorded, so the migration has genuinely nothing to do for it.

Nothing existing changes. Current subscribers keep billing their legacy Stripe Plan until #800 runs.

## What changed

**`SubscriptionItem.objects.resolve_purchase(plan, nonprofit)`** — returns `(canonical_plan, plan_price)` for a plan a customer picked. The plan they pick is not always the plan they end up on: annual and nonprofit are separate `Plan` rows today, and both collapse onto a canonical tier where the difference is carried by the price's `interval` and `label`.

**`plan_mapping.py`** now covers the ten purchasable slugs as well as the legacy ones. One table drives both the sign-up flow and the migration, so a new subscription is recorded *exactly* as a migrated one is. `resolve_target(allow_comped=False)` is what stops a self-service purchase resolving to a comped price.

**The billing shape follows the resolved price, not `plan.annual`.** Trusting the flag would let an annual price land on a subscription recorded as monthly — which groups it onto the wrong invoice and picks the wrong collection method. This is the subtle one.

**`get_plan_data()` reads the nonprofit rate from `PlanPrice`** rather than string-substituting the slug and reading `base_price` off a separate `sunlight-nonprofit-*` `Plan`. Display and billing now resolve the same way, so the figure shown is the figure sold — and it unblocks archiving those rows in #806. Falls back to the variant row while no `PlanPrice` exists.

**`add_subscription()` takes `nonprofit`**, threaded through from the form's checkbox.

## Degrades safely

`resolve_purchase` returns `(plan, None)` when there's nothing to resolve to — which is every plan until `consolidate_stripe_products` has run on production, and any plan the mapping doesn't cover. That leaves the subscription on the legacy plan and legacy Stripe id, exactly as it would have been. So this is safe to ship before that command runs; it just does nothing useful yet.

## QA

**Before the consolidation command has run**, everything below should behave exactly as it does on master — that's the fallback path, and it's worth confirming first.

**After `consolidate_stripe_products` has run on the preview deploy:**

- [x] **Buy an Organization plan (monthly).** The new `SubscriptionItem` has a `plan_price` at `organization / monthly / standard`, and `plan` points at the canonical Organization row. Check Stripe: the subscription bills the new **Price** id, not `squarelet_plan_organization`.
- [ ] **…and it bills one unit.** In Stripe the subscription item is **quantity 1** and the amount is the list price — $100 for Organization, not $500. Found 2026-09-21: the views passed `plan.minimum_users` as the quantity, which the legacy tiered Plans priced as the base and a flat Price prices as five of the tier. Production's canonical `organization` row has `minimum_users = 5`, so check this on a plan whose minimum is above 1; staging's rows may not be. The line in the admin shows quantity 1 too.
- [x] **Buy an annual plan.** The parent `Subscription.interval` is `annual` — *not* monthly. This is the bug the third bullet above prevents; if it's wrong the line lands on the wrong invoice group.
- [x] **Buy a Sunlight plan with the nonprofit box ticked.** The price recorded is `label="nonprofit"`, and the amount charged is the nonprofit rate.
- [x] **The Sunlight plan page shows the nonprofit price**, and the number matches what ticking the box actually charges.
- [x] **Buy two plans of the same interval for one organization.** Both lines land on one subscription and one invoice (that's #784's behaviour, worth re-confirming here).
- [x] **Buy an annual and a monthly plan for one organization.** Two separate subscriptions, two invoices — Stripe requires one interval per subscription.
- [x] **A comped organization is unaffected** — no self-service purchase should ever resolve to a `comped` price.

**Worth checking in the admin afterwards:** every subscription created during QA has a non-null `plan_price`. That is the property #800 depends on.


squarelet/organizations/models/organization.py
```diff
@@ -543,29 +543,44 @@ def member_users(self, request=None):
...
         plan, cancelled or not.  A cancelled line still occupies it:
         `unique_together` is (subscription, plan), so a second line for the
         same plan on the same subscription cannot exist.  Reviving one is
         `uncancel`'s job, reached through Resubscribe.
+
+        The question is asked about the *resolved* plan, because that is
+        what the line will be stored under.  Asking about the picked row
+        instead found nothing for anyone buying an annual or nonprofit
+        variant, and the check that exists to raise this error politely was
+        skipped in favour of an IntegrityError from inside the transaction.
```
eyeseast: I don't understand this comment at all. What question? Who's asking? There's some context missing here, or this is just Claude keeping notes for itself.

squarelet/organizations/models/organization.py
```diff
@@ -607,7 +627,7 @@ def add_subscription(self, plan, max_users, user, token=None, payment_method=Non
             user=user,
             reason=ChangeLogReason.updated,
             to_plan=plan,
-            to_max_users=max_users,
+            to_max_users=item.quantity,
```
eyeseast: Is `max_users` still going away?

mitchelljkotler:
Split into #833 → #834 → #835 → #836 (stacked on #831), rewritten for the release 3 subscription kinds and the terse comment rules. The `stripe_modify(proration_behavior=…)` override moves to #800, its only user. Original kept at tag `backup/split798/signup-price` locally and on this branch.

PR #803, opened 2026-09-03
Convert stripe html receipts to pdfs using weasyprint

mitchelljkotler:


duckduckgrayduck (approved):
Tested a subscription only on Accounts & purchasing requests on Muckrock both produce valid receipts.

PR #804, opened 2026-09-04
Redis fix

duckduckgrayduck:
We recently updated redis and our REDIS url changed. 
I was trying to figure out why we were getting SSL EOF errors coming from the Redis result store backend, like so: 
https://muckrock.sentry.io/issues/7659227976/?environment=production&project=1422309&query=trace%3A7eed3fd79b8a4df9892f725d3d7c4575%20%21issue.id%3A7710463416&referrer=issue-details-trace-preview-issue-stream
which propagates up to a Celery redis result store retry limit:
https://muckrock.sentry.io/issues/7710463416/?environment=production&project=1422309&query=is%3Aunresolved&referrer=issue-stream

Reading the docs, I see "as of redis-py 3.0 changes the default value of the ssl_cert_reqs option from None to ‘required’"
further 
"This check can be disabled by setting ssl_cert_reqs to None"
from https://pypi.org/project/redis/3.5.3/

Reading Heroku's documents, it looks like we MUST set ssl_cert_reqs to None 
https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-in-python
because they self sign their certificate. 

Ways to replicate this issue in the shell:

>>> import os, ssl, redis
>>> url = os.environ["REDIS_URL"]
>>> redis.from_url(url).ping()   
This raises. 

When I do 
redis.from_url(url, ssl_cert_reqs=ssl.CERT_NONE).ping()
It returns True.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -6,6 +6,7 @@ Base settings to build other settings files upon.
 from celery.schedules import crontab
 
 # Standard Library
+import ssl
 from datetime import timedelta
 
 # Third Party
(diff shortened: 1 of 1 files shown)


duckduckgrayduck:
This was already implemented in the Celery App code. It looks like Heroku's might be having  intermittent outages that I will need to follow up on with them about that is causing the SSL handshake drop outs

PR #808, opened 2026-09-08
Refactor and clean up card information displays

allanlasser:
I thought that I really landed a solid sectional information display in my most recent work on consolidated benefits. I took the patterns from those changes and applied them anywhere we were grouping information inside of cards: org memberships (both directions), invitations, and similar information was all using bespoke styling rules. This unifies it all into a `card.css` file and class-convention: `card-section`, `card-section-label`, `card-subsection`, `card-list` and `card-item`.

Impacted views:

- user detail (/users/x)
- organization detail  (/orgs/x)
- organization list (/orgs)
- member management (/orgs/x/manage-members)
- affiliate management (/orgs/x/manage-member-orgs/)
- invitation management (/orgs/x/invitations/, /orgs/x/requests/, /orgs/x/invitation/)


frontend/css/card.css
```diff
@@ -0,0 +1,111 @@
...
+  gap: 0.25em 0.5em;
+}
+
+.card-subsection .card-item {
+  border-top: 1px solid transparent;
+  border-bottom: 1px solid transparent;
+  transition: border-color 0.1s linear;
+}
+
+.card-subsection .card-item:hover {
```
dnass, line 73: On a two-item list, or the middle item of a three-item list, this border effect doesn't focus attention on the hovered item — it just looks like dividers are appearing for the whole list. A light background color on hover would fix the ambiguity, and would also work on `card-item`s that aren't in a subsection.

squarelet/templates/organizations/includes/plan_card.html
```diff
@@ -20,8 +20,8 @@
 {% if subscriptions or upgrade_plan or not inherited_benefits %}
 <main class="plan-card{% if not subscriptions %} plan-card--no-subscription{% endif %}">
```
dnass, line 21: This seems like a good opportunity to improve the semantic structure of this markup. A document [should only have one `main`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/main), so this can just be a div because it's always rendered inside a `section`. And divs with the `card-section` class should be `section`s because they contain `header`s.

dnass: There are a bunch of other `main`s that should be divs throughout these templates. If it's out of scope for this PR, let me know and I can open a separate issue.

allanlasser: This feels in scope to me and is great feedback!

dnass (approved):

PR #809, opened 2026-09-08
Step 1b — the Subscription / SubscriptionItem data model

mitchelljkotler:
Stacks on #781. **Split out of #784** so the structural change and the behaviour change can be read separately.

> ## ⚠️ Do not merge this to master on its own
>
> CI/CD deploys automatically on merge to master, and this half has the columns but not the behaviour that keeps billing pages, renewal dates and the reconciliation commands correct. **Merge #784 into this branch first, then merge this to master** — the two are one deployable unit.
>
> Everything above #784 in the stack stays individually deployable; this pair is the exception.

## What this is

The rename and the restructure, and nothing that changes what the system *does*. The question to read it with is **"does this still mean the same thing?"** — almost every hunk is the same expression reached a different way.

- `Subscription` → `SubscriptionItem`, and a new `Subscription` parent that owns the Stripe subscription. Stripe requires every item on a subscription to share a billing interval and collection method, so an organization needs one subscription per combination it holds — that's what the uniqueness constraint encodes.
- Subscription-level state (status, period end, cancellation) moves to the parent, so a renewal webhook updates one row rather than fanning out across items that could then disagree.
- 50 files of mechanical repointing.

## What it deliberately isn't

Per-item cancellation, the Stripe sync logic, the billing pages, the renewal date moving off the API, the reconciliation commands — all in **#784**, which is 25 files and where every review finding so far has landed.

## Not independently functional

**Earlier revisions of this description said "there should be nothing to test behaviourally." That was wrong**, and worth correcting rather than quietly fixing: on this branch alone, `get_subscription_next_date`, `BaseResubscribe.post`, `audit_subscriptions`, `sync_subscriptions`, and the renewal-date text on the org and user pages are all broken — the parent model exists here, but the code that keeps it correct after a plan change is #784's. Every one of those is repaired there.

Nothing that reaches master is broken, because the two merge together — but reviewing this branch in isolation, or deploying it on its own, will surface real breakage that is not a bug in this PR so much as work that hasn't landed yet.

## Migrations

**These are the ones that ship.** They used to be finalised by the PR above — three migrations here, squashed to two up there — so this section told you to go and read them somewhere else. They have been moved down, and this branch now carries the final form.

`0083` renames, `0084` adds the parent and moves subscription-level state onto it, grouping existing lines by `(organization, interval, collection_method)` rather than assuming one per organization. The new uniqueness constraint is one parent per billing shape, and a per-item parent would violate it the moment an organization held two lines of the same shape. Production confirms none does today; it is written to hold regardless.

It refuses rather than guesses in two cases, neither of which can fire on today's data:

- **Two lines of one shape naming different Stripe subscriptions.** Merging them would leave one billing with nothing pointing at it, so it names both and stops.
- **Lines of one group disagreeing about cancelling.** The parent holds one answer and there is nowhere to put the other until per-line cancellation lands in #784. Collapsing with `all` renews a plan the customer cancelled; `any` cancels ones they kept.

**This release rolls back.** `0083` relaxes `organization` to nullable before the data move so reversing can repopulate it, and a migration test reverses with a real line in the table and checks the organization, the Stripe id and the status all come back.

## QA

**There is nothing to exercise on a review app.** This branch is not independently functional — the renewal date on the organization and user pages, `BaseResubscribe.post`, `audit_subscriptions` and `sync_subscriptions` are all broken until #811, which is why the two merge as one unit. Anything you find on a preview deploy of *this* branch is expected and fixed in #811; QA the pair from #811, not from here.

What is worth doing here:

- [ ] Read the data migration. It is the part that cannot be undone by a later commit.
- [ ] Run it against a staging copy of production and confirm it completes without the "more than one Stripe subscription" refusal firing anywhere real.


eyeseast (approved):

allanlasser (approved):

allanlasser:
If we can't merge this onto `master` directly, we should create a release branch and merge into that.

PR #811, opened 2026-09-10
Split Subscription from SubscriptionItem

mitchelljkotler:
Stacks on #781. **Split out of #784**, which is now the behaviour change alone.

## What this is

`Subscription` becomes two models: a parent owning the Stripe subscription, and `SubscriptionItem` lines billed on it. Stripe requires every item on a subscription to share a billing interval and a collection method, so an organization needs one subscription per combination it holds — that is what the new uniqueness constraint encodes, and what lets several plans arrive on one invoice.

**Behaviour is master's.** Cancellation stays a property of the subscription: `SubscriptionItem.cancelled` and `cancel_at` are properties reading the parent's, so templates and the admin are unchanged, and cancelling any line still ends the whole subscription the way the button always has. Cancelling one plan of several is **#784**.

That is the point of the split: this half can be deployed and watched on its own, and if something moves you know it was the restructure rather than the new feature.

## What is genuinely new

Correctness against Stripe, all of it found by running the code against a real sandbox rather than by reading it:

- **Lines are identified before they are described.** `stripe_item_id` starts empty on everything the migration touches, and a line spec with no id asks Stripe to *add* a line — so the first change to any existing subscription was rejected outright (`an existing Subscription Item is already using that Price`), and a plan change would have billed the customer for the plan they left as well as the one they chose.
- **`sync_to_stripe`** covers the three transitions a plan change can cause — becoming free, becoming paid, staying paid. `stripe_modify` alone silently no-ops without a Stripe subscription, so an upgrade off free granted paid access and never billed.
- **The renewal date is read from the database**, not from a live Stripe call per line on every page render.
- **The reconciliation commands and the audit** understand lines.

## New: a Stripe sandbox test suite

`squarelet/organizations/tests/test_stripe_sandbox.py` — 17 tests that run against a **real Stripe sandbox** rather than a mock. This is new test infrastructure for the repo, not just tests for this PR.

**Every correctness fix listed above was found by running it.** None came from reading the code. The suite was written after the split had already been reviewed once, and on its first run it failed three of the fixes that had been made in response to that review — they had been verified against mocks that happily accepted calls Stripe rejects. Mocks cannot tell you that `StripeObject` has no `.get()`, that `sub.items` collides with a dict method, or that a subscription has quietly grown a second line billing the plan the customer just left.

**It is opt-in and offline by default.** `pytest.ini` adds `-m "not stripe"`, so `inv test` stays hermetic and nothing in CI reaches the network. `inv test-stripe` runs them. They need `STRIPE_SANDBOX_SECRET_KEY` in the Django env file — Stripe sandboxes are free and isolated, so this is its own account rather than shared test mode — and the fixture refuses to run against a key starting `sk_live`.

**What it covers:** what the customer is actually charged (legacy group plans bill `tiered`/`graduated`, where quantity *selects* a tier rather than multiplying — reading that wrong overcharges a five-seat subscriber fivefold); line identity across add, remove and plan change; collection method for card versus invoiced payers; cancel/resubscribe round-trips; and the `backfill_stripe_item_ids` command against subscriptions Stripe actually holds.

Worth running on any later PR in this stack that touches billing — it has caught something every time the Stripe-facing code moved.

## Migrations

**None.** They live in #809, in the form that actually deploys — this branch used to rewrite them, which meant #809 was reviewed against an intermediate set that never reached master. Moved down, so each PR is reviewed on what it ships.

## QA

- [ ] **Run `backfill_stripe_item_ids` after deploying.** Without it the first plan change on any existing subscription is rejected by Stripe. `--dry-run` lists what it would fill in.
- [ ] **Add a plan to an organization that already has one.** They keep one Stripe subscription with two line items, and no duplicate of the plan they already had. A card payer is invoiced for the proration straight away — a *second* invoice, covering the rest of the current period only, not a full month. An organization billed by invoice gets no new invoice; the charge appears on their next scheduled one, which you can see on the customer's upcoming invoice in Stripe.
- [ ] **Cancel an organization holding two plans, and expect *both* to end.** This is master's behaviour, preserved deliberately, and it is the thing most likely to look like a bug during QA: cancellation lives on the subscription, `SubscriptionItem.cancelled` is a property reading the parent's, and Stripe has no per-item `cancel_at_period_end`. So cancelling one line shows "Ends on" against every line, and resubscribing revives every line. **Per-line cancellation is #784, the next PR in the stack** — do not raise it against this one.
- [ ] **Cancel a subscription.** It shows "Ends on", keeps working until the period ends, and Stripe shows `cancel_at_period_end`.
- [ ] **Resubscribe** clears it, on Stripe as well as locally.
- [ ] **Upgrade an organization from a free plan to a paid one** — a Stripe subscription is created and the first invoice is charged.
- [ ] **The organization and user pages show the right renewal date**, without a Stripe call per line.
- [ ] **Run `audit_subscriptions`** and confirm it is as clean as before the deploy. A paid line reporting `missing on stripe` with `(no stripe_item_id)` means the backfill did not reach it.

Most of the above is already covered by `inv test-stripe` against a real sandbox — see the section above. Run it before merging if you touch anything Stripe-facing.


mitchelljkotler:
I ran through the QA check list on the review app

duckduckgrayduck:
I'm receiving OperationalErrors in the deploy preview environment related to handle_charge_succeeded and the stripe webhook. It's probably deploy preview specific, as I never get these on production and occasionally do on previews, nonetheless I wanted to note it. 
https://muckrock.sentry.io/issues/7735695280/?project=1247105&query=is%3Aunresolved&referrer=issue-stream
https://muckrock.sentry.io/issues/7738646633/?project=1247105&query=is%3Aunresolved&referrer=issue-stream

When I refresh, it does tell me I'm subscribed with the professional plan. 
When I go to view my receipt, it is the old style receipt not the Stripe receipt. This is a regression that needs fixed. 

I also get no email that I've subscribed, which isn't something I've noticed yet on production. Looks like this is an issue on staging too, not sure when that broke. 

Changing the current period end and cancel_at date in the admins make no affect on Stripe, it remains Oct 17th and the subscription remains active. Probably want to make them read only in the admin to prevent someone from mucking about. This is caught by the audit_subscriptions management command though, which is nice.

eyeseast (approved):

eyeseast:
## Review findings

Reviewed at `b1f8cb1d`, in the context of the full stack. Scoped to what is actionable on this branch — the `.paid()`/`is_free` split and the `0084` `collection_method` inference belong to later branches and #809 respectively, and `postdeploy.sh` is a stack-wide deploy concern.

### A line that goes free is never removed from Stripe — `payment.py:577`

`stripe_items()` skips free lines, and `stripe_modify()` passes that list to `Subscription.modify(items=...)`. Stripe leaves items *omitted* from that list untouched — removal requires an explicit `{"id": ..., "deleted": True}` spec.

Scenario: a subscription holds two paid lines A and B; `SubscriptionItem.modify()` moves A onto a free plan. `sync_to_stripe()` sees `self.free == False` (B is still paid), so it takes the `stripe_modify()` branch and sends only B's spec. A's Stripe item survives and the customer keeps being charged for the plan they downgraded off.

Still present at the tip of the stack, so it will not be resolved downstream. The sandbox suite covers only the *last* paid line going free (`test_moving_to_a_free_plan_stops_the_billing`), which takes the delete branch instead.

### `auto_renew` `all()` → `any()` — `payment.py:554`

The docstring's premise — "that line stops on its own, through `cancelled`/`cancel_at`" — is not true on this branch. `SubscriptionItem.cancelled` is still a read-through property of the parent; per-line cancellation arrives in #784.

Scenario: an org holds a renewing monthly plan and a `Plan.auto_renew=False` plan on the same subscription. Previously `cancel_at_period_end=True` was sent and the non-renewing plan stopped. Now `any()` is True, `cancel_at_period_end=False` goes to Stripe, and a plan explicitly configured not to renew (see `admin.py:664`) bills indefinitely.

The PR body argues this half is separately deployable. That holds for everything else here, but not for this change — it is only safe once #784 lands. Worth deciding explicitly: defer the change to #784, or accept that #811 must not deploy without it.

### `refresh_from_db()` discards the caller's unsaved fields — `payment.py:1211`

`SubscriptionItem.modify()` calls `self.refresh_from_db()` before `self.plan = plan; self.save()`. Its documented caller, `Organization.modify_subscription` (`organization.py:713-714`), does `sub.quantity = max_users` and then `sub.modify(new_plan)` — so the new seat count is silently reverted both locally and on Stripe, while the ChangeLog row records the new `to_max_users`. Reach is narrow (no view or API caller), but the audit trail ends up lying. Re-apply the in-memory values after the refresh, or refresh only `stripe_item_id`.

### A new line can join a cancelled subscription — `querysets.py:442`

`Subscription.objects.get_or_create(organization, interval, collection_method)` will return a subscription with `cancelled=True`.

Scenario: an org cancels plan A (subscription `cancelled=True`, `cancel_at` set), then buys plan B with the same interval and collection method. B joins the cancelled subscription; `stripe_modify` sends `cancel_at_period_end=self.cancelled or ...` → True (pinned by `test_a_pending_cancellation_survives_adding_a_plan`), and `proration_behavior="always_invoice"` charges a card payer immediately — for a plan that `restore_organization` (`tasks.py:59-63`) will delete at `cancel_at`.

`keep_renewing_for` fixes this at the tip of the stack, so it self-resolves if the stack always ships together. It is a live bug in any window where this branch is merged and the later ones are not.

### Smaller

- **`payment.py:900`** — the `sync_to_stripe` free branch clears `subscription_id` and the cancellation pair but leaves `stripe_status` and `current_period_end`. After a downgrade to free, `Subscription.next_date` still returns a date, so `plan_card.html` renders "Renews &lt;date&gt;" for a subscription that no longer exists on Stripe.
- **`backfill_stripe_item_ids.py:511`** — the queryset selects subscriptions where a line lacks an id *or* `current_period_end` is null, but `_report()` only inspects missing line ids and returns `"skipped"` otherwise. A subscription visited solely to cache its renewal date reports "nothing to do" under `--dry-run` and is then written on the real run. The QA checklist tells operators to trust `--dry-run`.
- **`squarelet/templates/organizations/plan_info.html`** — not `{% include %}`d by any template; grep for `_cls-planInfo` / "Current plan" finds only this file and its test. The summary block it was extracted from no longer renders anywhere, while a test asserts that it does.

### Adjacent, not introduced here

`_reconcile_cancelled_subscription` (`tasks.py:678`) does a bare `subscription.delete()` and is unchanged by this PR, so it is not a defect of this branch. But this branch is what makes it dangerous: one `Subscription` row now carries many lines, and that delete cascades to all of them. It runs from `handle_subscription_deleted` and from `handle_subscription_updated` on `status == "canceled"` — the normal end-of-life path, which Stripe fires at period end, usually before the nightly sweep. Worth addressing here rather than booking it as pre-existing.

### Checked and cleared

`always_invoice` on a no-op modify, `settle_added_line` on `send_invoice` orgs, and the free-line-added-to-paid-subscription path are all pinned by `test_stripe_sandbox.py`. The `transaction.atomic()` added in `querysets.start()` does not change rollback semantics — `Organization.add_subscription` was already `@transaction.atomic` on master.

squarelet/payments/views.py
```diff
@@ -556,9 +553,13 @@ def get_context_data(self, **kwargs):
...
         # Get subscriptions and add renewal/cancellation date and cost data
-        subscriptions = self.object.subscription_items.all()
+        subscriptions = self.object.subscription_items.select_related(
+            "subscription", "plan"
+        )
         for subscription in subscriptions:
-            subscription.next_date = get_subscription_next_date(subscription)
+            # `next_date` is a property on the line now, reading its
+            # subscription's - assigning to it raises, and this page is the
+            # only caller that did.
```
allanlasser: This is a noisy comment that now states the obvious; we don't need to explain a caller / assignment behavior in this context when it's been removed!

allanlasser (changes requested):
I'm finding that the verbose comments are making review very challenging.

The comments themselves are hard to read. They're written in Claude's characteristic voice with paragraphs of long, run-on sentences. In most cases, a simple sentence or two would be sufficient. My mind starts to glaze over as I'm reading a chunk of prose that I struggle to make sense of.

I'm not finding that they contribute to understanding what's happening in the code, they seem more focused on describing the thought process behind changes. This also ends up describing why code that was removed no longer exists—not super helpful (see inline comment below).

For similar reasons, I am finding it very hard to understand how this piece fits into the big picture from the PR description.

---

Here's an example of what I'm talking about, from the PR description:

> Behaviour is master's. Cancellation stays a property of the subscription: `SubscriptionItem.cancelled` and `cancel_at` are properties reading the parent's, so templates and the admin are unchanged, and cancelling any line still ends the whole subscription the way the button always has. Cancelling one plan of several is https://github.com/MuckRock/squarelet/pull/784.
>
> That is the point of the split: this half can be deployed and watched on its own, and if something moves you know it was the restructure rather than the new feature.

What are you trying to say here? I'm interpreting it as, "We shouldn't observe any behavior changes here, this is a restructuring of our existing logic." But "cancelling any line still ends the whole subscription the way the button always has" is a really obtuse expression of that idea.

---

I know that this has been blocked waiting for review for a few weeks, and I think the comment style here is contributing to that slowdown.

I think this would benefit from updating our `CLAUDE.md` to include some style guidance for code comments / PR descriptions, then updating this and other PRs in the stack to conform to it.

allanlasser:
As far as breaking down the complexity of this PR, separating the "Stripe sandbox test suite" and the resulting fixes would really help understand what is specifically changing around the Subscription / SubscriptionItem logic split.

mitchelljkotler:
> I'm finding that the verbose comments are making reviewing this very challenging for me.

Yes, this is a very good point.  I have added a CLAUDE.md and edited down the comments heavily.  Let me know if this is reviewable.  I can break out the Stripe tests if needed, but would prefer to not have to do git surgery and get this PR released this week.  I am going to look into breaking down the following PRs into more manageable sizes.

.gitignore
```diff
@@ -165,7 +169,6 @@ typings/
 !.vscode/extensions.json
 
 ### Claude Code
-CLAUDE.md
```
allanlasser: Just flagging a risk to other contributors—any local CLAUDE.md will conflict with this canonical version; may require further reconciliation of rules, but the file in this PR is a great starting point.

squarelet/organizations/querysets.py
```diff
@@ -439,22 +440,37 @@ def start(self, organization, plan, payment_method="card", quantity=1):
...
+                    anchor_day=anchor.day if anchor else None,
+                )
+            else:
+                # The cached object was fetched before the line was added.
+                stripe_subscription = subscription.stripe_modify()
+                if stripe_subscription is not None:
+                    subscription.settle_added_line(stripe_subscription)
+
+        # `notify_started` gates on entitlement; gating on price here would
+        # drop free signups that carry `organization`.
```
allanlasser: In this case, "free signups" refer to comped plans with a $0 price, correct?

mitchelljkotler: At this point, we are not using PlanPrice yet, so it is any plan with a $0 price, which could be something like an Admin or Beta plan, or a custom plan with $0 price.

allanlasser (approved):

PR #813, opened 2026-09-11
Sync client stats to Accounts and display them in the admin. 

duckduckgrayduck:
- Adds client stats fields to user and org models
- Adds a management command and nightly celery task that traverses both client stats apis fetching stats and merging them in. 
- Displays these user stats for now in the admin. 

Requires https://github.com/MuckRock/muckrock/pull/2279 for MuckRock's side to work and github.com/MuckRock/documentcloud/pull/465 for DocumentCloud stats api optimizations to be made to succeed. 

A follow up with displaying them on user/org pages for users to view about themselves/their org can be made afterward. After this, a follow up PR will add these to the Odoo sync.


duckduckgrayduck:
The MuckRock sync worked with page sizes of 500 specified and ran quickly and without any issues. DocumentCloud's page size is capped to 100 (it shares pagination with the main API which is capped at 100 max page size) but given the current and predicted future user counts, this should have no problem finishing in under half the allotted task time. If we split the page size setting for the stats api in DocumentCloud's backend in the future, we can bump the page size to 500 and it should be quicker. Not a blocker however. You can view sample stats on your own user or my user at the Deploy preview in the admin panel.

squarelet/organizations/admin.py
```diff
@@ -558,6 +560,13 @@ def user_link(self, obj):
 
     user_link.short_description = "User"
 
+    def client_stats_display(self, obj):
+        """Read-only dump of the stats synced from client services."""
+        json_data = json.dumps(obj.client_stats, indent=4)
+        return mark_safe(f"<pre>{json_data}</pre>")
```
mitchelljkotler: Safer as :
`return format_html("<pre>{}</pre>", json_data)`
in case there is any html in the json_data

squarelet/users/admin.py
```diff
@@ -247,6 +256,13 @@ def all_org_links(self, obj):
 
     all_org_links.short_description = "All Organizations"
 
+    def client_stats_display(self, obj):
+        """Read-only dump of the stats synced from client services."""
+        json_data = json.dumps(obj.client_stats, indent=4)
+        return mark_safe(f"<pre>{json_data}</pre>")
```
mitchelljkotler: Safer as :
return format_html("<pre>{}</pre>", json_data)
in case there is any html in the json_data

mitchelljkotler (changes requested):

eyeseast (approved):

